PR #1, opened 2020-10-31
Set up workflow to check license and lint codes

kezhenxu94:


.golangci.yml
```diff
@@ -0,0 +1,96 @@
...
+      -
+    include-go-root: true
+    packages-with-error-messages:
+      fmt: "logging is allowed only by logutils.Log"
+  misspell:
+    locale: US
+  lll:
+    line-length: 150
+  goimports:
+    local-prefixes: github.com/apache/skywalking-cli
```
wu-sheng, line 43: copy/paste issue?

kezhenxu94: yes, fixed

wu-sheng (approved):

PR #8, opened 2020-12-08
Add main structure

EvanLjp:
enhance the plugin mechanism, please look at the design doc


wu-sheng:
First, please fix Makefile, `make all` doesn't work, due to no `build` task.

internal/pkg/plugin/definition.go
```diff
@@ -0,0 +1,35 @@
...
+	Description() string
+}
+
+// NameFinderFunc is used to get the plugin name from different plugin configs.
+type NameFinderFunc func(config interface{}) string
+
+// InitializingFunc is used to initialize the specific plugin.
+type InitializingFunc func(plugin Plugin, config interface{})
+
+// CallbackFunc would be invoked after initializing.
```
surlymo, line 34: why split it from initializingfunc？show some use cases?

plugins/fallbacker/timer/timer_fallbacker.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"time"
+
+	"github.com/apache/skywalking-satellite/internal/pkg/event"
+	"github.com/apache/skywalking-satellite/plugins/forwarder/api"
+)
+
+// Fallbacker is a timer fallbacker when forward fails. `latencyFactor` is the standard retry duration,
+// and the time for each retry is expanded by 2 times until the number of retries reaches the maximum.
+type Fallbacker struct {
```
surlymo, line 29: i think fallbacker should be a plugin?

plugins/fallbacker/timer/timer_fallbacker.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (t *Fallbacker) DefaultConfig() string {
+	return `
+max_times: 3
+latency_factor: 2000
+`
+}
+
+func (t *Fallbacker) FallBack(batch event.BatchEvents, connection interface{}, forward api.ForwardFunc) {
```
surlymo, line 49: do not make func name the same as type name

plugins/fallbacker/timer/timer_fallbacker.go
EvanLjp: yes it is a plugin, api/xxx is interface , this is on timer package

internal/pkg/plugin/definition.go
EvanLjp: the module and the extension plugin has different initial ways

docs/design/plugin_structure.md
```diff
@@ -0,0 +1,47 @@
...
+2. Find plugin type according to the plugin category(interface) and the plugin name.
+3. Create an empty plugin.
+4. Initialize the plugin according to the config.
+5. do some callback processing after initialized.
+
+
+## Plugin usage in Satellite
+Not only does the extension mechanism depend on Plugin, but the core modules also depend on Plugin in Satellite. We'll illustrate this with an example.
+
+- Core Module interface: module.Service
```
surlymo, line 32: it seems simpler if merging defaultplugin & service design. plugin only have four layers:
plugin basic init -> lifecycle management -> specific plugin interface -> specific plugin implements

plugins/collector/api/collector.go
```diff
@@ -27,7 +27,7 @@ import (
 //   Init()     Initial stage: Init plugin by config
 //    ||
 //    \/
-//   Prepare()   Preparing stage: Prepare the collector, such as build connection with SkyWalking javaagent.
+//   Init()   Preparing stage: Init the collector, such as build connection with SkyWalking javaagent.
```
wu-sheng, line 30: Why duplicated `Init`? And `such as build connection with SkyWalking javaagent.` is wrong. I think you mean, set up the gRPC server for the agents?

EvanLjp: it is a mistake when use goland refactor

plugins/fallbacker/api/fallbacker.go
```diff
@@ -22,23 +22,24 @@ import (
...
+	"github.com/apache/skywalking-satellite/plugins/forwarder/api"
 )
 
 // Fallbacker is a plugin interface, that defines some fallback strategies.
 type Fallbacker interface {
 	plugin.Plugin
-
 	//  FallBack returns nil when finishing a successful process and returns a new Fallbacker when failure.
-	FallBack(batch event.BatchEvents) Fallbacker
+	FallBack(batch event.BatchEvents, connection interface{}, forward api.ForwardFunc, callback DisconnectionCallback) Fallbacker
```
wu-sheng, line 32: Always prefer `code` rather than `nil` to determine the behavior.

internal/pkg/plugin/registry.go
```diff
@@ -18,62 +18,85 @@
...
 
-// Add new plugin category. The different plugin category could have same plugin names.
-func AddPluginCategory(pluginCategory reflect.Type) {
+// RegisterCategory register new plugin category with default InitializingFunc.
+// required:
+// pluginCategory: the plugin interface type.
+// Optional:
+// n: the plugin name finder,and the default value is defaultNameFinder
+// i, the plugin initializer, and the default value is defaultInitializing
+// c, the plugin initializer callback func, and the default value is defaultCallBack
```
wu-sheng, line 45: Unmatched comments.

internal/pkg/plugin/registry.go
```diff
@@ -18,62 +18,85 @@
...
 	"sync"
 )
 
-// All plugins is wrote in ./plugins dir. The plugin type would be as the next level dirs,
-// such as collector, client, or queue. And the 3rd level is the plugin name, that is also
-// used as key in pluginRegistry.
-
-// reg is the global plugin registry
+// the global plugin registry
 var (
```
wu-sheng, line 28: Is this really required to be global? Take a look at Java version's `ModuleManager`. Once you accept the global singleton, the tests are sensitive about clearing this register table after every time, and you don't have a chance to run tests concurrently(even I don't know whether this is an option in go).

plugins/collector/api/collector.go
```diff
@@ -24,36 +24,22 @@ import (
...
-//    \/
-//   Close()    Closing stage: Close the Collector, such as close connection with SkyWalking javaagent.
-
 // Collector is a plugin interface, that defines new collectors.
 type Collector interface {
 	plugin.Plugin
-
-	// Prepare creates a listen or reader to gather data.
-	Prepare()
+	// Prepare creates a listener or reader to gather APM data, such as build connection with SkyWalking javaagent.
```
wu-sheng, line 30: `such as build connection with SkyWalking javaagent.` still a wrong statement, you can't connect to javaagent.

internal/pkg/plugin/registry.go
EvanLjp: It should be global because the registered way is func init()

wu-sheng: This leads to another question, why needs the `package init()` mechanism? All the plugins should be registered due to the configuration files or parameters, aren't they? Otherwise, you will cost the memory for nothing.

wu-sheng: Golang app has nothing like SPI in Java, a hardcode `if/else` section could take care of the register easily. Better to read, and easy to extend, no extra memory cost to maintain this register table.

EvanLjp: the table only registers the reflect.type, I don't think this would cost resources for future extensions.

EvanLjp: A similar SPI mechanism is implemented, I don't think hard code to register plugins is a good way, Similar patterns exist in many places, such as mosn, beats, and etc... . In the original version, I also experimented with the golang dynamic link(.so mechanism), but it was difficult to use

EvanLjp: there is a demo to compute the size of (reflect.type), the size is only few Bytes, https://stackoverflow.com/questions/31338588/generic-function-to-get-size-of-any-structure-in-golang

wu-sheng: I know you are writing an SPI mechanism in the structure. Personally, I have concerns about that. 
This thing makes the codes more complex than it needs to be, at the same time, w/o the dynamic link(I don't ask for implementing dynamic link) this mechanism provides no new feature.
This is the reason mostly.

Also, around the tech part
> In Go, the predefined init() function sets off a piece of code to run before any other part of your package. This code will execute as soon as the package is imported.

How could you sure the package is imported if the implementation in a random package? Or some packages the core doesn't know to access in the initializing stage?

___
In the basic idea, I don't want the go works like Java. @hanahmily @kezhenxu94 I hope to get your inputs.

EvanLjp: if a specific plugin want to register itself to the registry, it must invoke the register method in its init func. That purpose the specific plugin must import the registry package. In the golang, the import package init would be first invoked. The reason why i want to use init func is it is hard to control the used plugins by manual registration.

EvanLjp: Satellite supports multi plugins ，so we support a registration mechanism to control this plugins. So there are 2 ways to register the plugins, which are auto registration and manual registration. Auto registration is relay on the init func in golang, that makes the codes more cleaner. However, it also makes the codes more complex. Another ways is manually registration, that would add few go files to control the plugins registration. This way would easier to read source code, but would add some files to control the registration. Which one should we choose？

wu-sheng: Note, auto-registration would cost clear memory, even limited, our SLO is 20M for the whole satellite deployment, consider this in the decision.

wu-sheng: Please edit this comment to show your preference. @apache/skywalking-committers, especially reviewers, @surlymo @hanahmily @kezhenxu94 @dmsolr @zhangkewei @gxthrj @nic-chen @moonming 

- +1 for Auto-registration. 
- +1 for manual registration in go source codes. @wu-sheng

nic-chen: > This leads to another question, why needs the `package init()` mechanism? All the plugins should be registered due to the configuration files or parameters, aren't they? Otherwise, you will cost the memory for nothing.

+1

EvanLjp: ok, I would change to the manual way.

surlymo: +1 for manual registration in go source codes

i think either is ok. auto-registration like spring annotation declaration in java, manual registration like spring xml definetion. i prefer the latter.

docs/design/plugin_structure.md
```diff
@@ -0,0 +1,47 @@
...
+`Plugin is a common concept for Satellite. Not only does the extension mechanism depend on Plugin, but the core modules also depend on Plugin`
+
+## Registration mechanism
+
+The Plugin registration mechanism in Satellite is similar to the SPI registration mechanism of Java. 
+Plugin registration mechanism supports to register an interface and its implementation, that means different interfaces have different registration spaces.
+We can easily find the type of a specific plugin according to the interface and the plugin name and initialize it according to the type.
+
+structure:
+- code: map[reflect.Type]map[string]reflect.Value
```
gxthrj, line 11: `[]` is a markdown keyword that needs to be escaped

docs/design/plugin_structure.md
```diff
@@ -0,0 +1,47 @@
...
+Not only does the extension mechanism depend on Plugin, but the core modules also depend on Plugin in Satellite. We'll illustrate this with an example.
+
+- Core Module interface: module.Service
+    - the specific plugin: module.Gatherer
+    - the specific plugin: module.Processor
+    - the specific plugin: module.Sender
+    - the specific plugin: module.ClientManager
+- Extension: 
+    - Collector interface 
+        - the specific plugin: segment-receiver
```
gxthrj, line 39: Is the `fetcher` plugin also specified here?

plugins/collector/api/collector.go
```diff
@@ -24,36 +24,22 @@ import (
...
-//   Close()    Closing stage: Close the Collector, such as close connection with SkyWalking javaagent.
-
 // Collector is a plugin interface, that defines new collectors.
 type Collector interface {
 	plugin.Plugin
-
-	// Prepare creates a listen or reader to gather data.
-	Prepare()
+	// Prepare creates a listener or reader to gather APM data, such as start a gRPC listener for Segment receiving.
+	Prepare() error
```
gxthrj, line 31: If we need to implement a `http-receiver`, should we provide concrete implementation here?

gxthrj: And do we need a type field to identify the `receiver` protocol, such as `http` `grpc`.

EvanLjp: I don't think the upper layers need to care about protocols, they're just different plug-ins

docs/design/plugin_structure.md
EvanLjp: fetcher is a kind of collector.

wu-sheng:
The new plugin design makes me more confortable now. Thanks for the refactoring, @EvanLjp

internal/satellite/module/buffer/buffer.go
```diff
@@ -0,0 +1,94 @@
...
+		first: 0,
+		last:  0,
+		size:  0,
+		cap:   capacity,
+	}
+}
+
+// Buf returns the cached data in BatchBuffer.
+func (b *BatchBuffer) Buf() []*event.OutputEventContext {
+	b.Lock()
```
surlymo, line 50: so many locks. it may leads to the system slowing down ？

wu-sheng: Yes, I want to say this. Is this really necessary for using lock even in the register? Who will use concurrency-registration?

wu-sheng: Any update?

EvanLjp: I am polishing it now

plugins/collector/api/collector.go
EvanLjp: Take a look again. Related concepts have been added

EvanLjp:
@wu-sheng  @gxthrj @gxthrj refactoring(add server concept) & polish codes & add docs   please read doc first, thx

docs/design/module_design.md
```diff
@@ -0,0 +1,127 @@
+# Module Design
+## Namespace
+Namespace is an isolation concept in Satellite. Each namespace has one agent to collect the APM data.
+So many agents would be running at the same time in Satellite.
```
wu-sheng, line 4:
```suggestion
The namespace is an isolation concept in Satellite. Each namespace has one pipeline to process the telemetry data(metrics/traces/logs). Two namespaces are not sharing data.
```
I think you mean this, please update the doc accordingly.

docs/design/module_design.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Plugins
+
+Plugin is the minimal components in the module. There are 2 kinds of plugins in Satellite, which are sharing plugins and normal plugins.
+
+- a sharing plugin instance could be sharing with multiple modules in the different namespaces.
+- a normal plugin instance is only be used in a fixed module of the fixed namespaces.
+
+### Sharing plugin
+Nowadays, there are 2 sharing plugins in Satellite, which are server plugins and client plugins. The reason why they are sharing plugins is to reduce the resource cost in connection. Server plugins are sharing with the ReceiverGatherer modules in the different namespaces to receive the external requests. And the client plugins is sharing with the Sender modules in the different namespaces to connect with external services, such as Kafka and OAP.
```
wu-sheng, line 43: I think this is not `2 sharding plugins`. You mean there are 2 kinds of sharding plugins?

docs/design/module_design.md
```diff
@@ -0,0 +1,127 @@
...
+ --------------------------------------------------------------------
+ --------------------------------------------------------------------
+|       ------------------      -----------      --------            |
+|      | ReceiverGatherer | => | Processor | => | Sender |           |                          
+|       ------------------      -----------      --------            |
+ --------------------------------------------------------------------
+```
+
+### Normal plugin
+There are 7 normal plugins in Satellite, which are Receiver, Fetcher, Queue, Parser, Filter, Forwarder, and Fallbacker.
```
wu-sheng, line 65:
```suggestion
There are 7 kinds of normal plugins in Satellite, which are Receiver, Fetcher, Queue, Parser, Filter, Forwarder, and Fallbacker.
```
Same as above.

docs/design/module_design.md
wu-sheng: APM data should be telemetry data.

docs/design/module_structure.md
```diff
@@ -1,18 +1,25 @@
...
+- The Gatherer module is responsible for fetching or receiving data and pushing the data to Queue. So there are 2 kinds of Gatherer, which are ReceiverGatherer and FetcherGatherer.
+- The Processor module is responsible for reading data from the queue and processing data by a series of filter chains.
+- The Sender module is responsible for async processing and forwarding the data to the external services in the batch mode. After sending success, Sender would also acknowledge the offset of Queue in Gatherer.
 
-- ClientManager: Maintain connection and monitor connection status
-- Sender: Sender data to the external services, such as Kafka and OAP
-- Gatherer: Gather the APM data from the other systems, such as fetch prometheus metrics.
-- Processor: Data processing to create new metrics data.
+```
+                            Agent
```
wu-sheng, line 12: I think this should be `namespace`. Don't put 2 names in one concept.

docs/design/plugin_structure.md
```diff
@@ -27,20 +26,21 @@ Initialize processing is like the following.
 4. Initialize the plugin according to the merged config, which is created by the input config and the default config.
 
 
+
 ## Plugin usage in Satellite
+Nowadays, we have 9 kinds of plugins in Satellite. 2 kinds of Plugins are the [sharing plugins](module_design.md) and the others are the [normal plugins](module_design.md).
```
wu-sheng, line 31: The word `kinds` is over-used. Consider using `Catalog S` for sharding plugins and `Catalog N` for normal plugins?

docs/design/plugin_structure.md
```diff
@@ -27,20 +26,21 @@ Initialize processing is like the following.
...
-        - the specific plugin: gRpc-client
-    - Fallbacker interface
-        - the specific plugin: timer-fallbacker
-    - Forwarder interface
-        - the specific plugin: grpc-segment-forwarder
-    - Parser interface
-        - the specific plugin: csv-parser
+    - sharing plugins
+        - Server interface
+        - Client interface
```
wu-sheng, line 38: Let's don't use `interface`, this is not a design concept, this is the code. You called them as plugins in other parts of the doc.

configs/satellite_config.yaml
```diff
@@ -0,0 +1,69 @@
...
+sharing:
+  clients:
+    - plugin_name: "grpc-client"
+      k: v
+  servers:
+    - plugin_name: "grpc-server"
+      k: v
+agents:
+  - common_config:
+      namespace: namespace1
```
wu-sheng, line 49:
```suggestion
namespace:
  - common_config:
      name: namespace1
```

docs/design/module_structure.md
EvanLjp: Agent is a whole concept:   Each namespace has one agent to collect the telemetry data. (module_design.md)

EvanLjp: would fix it by using one concept

docs/design/module_design.md
```diff
@@ -0,0 +1,128 @@
...
+ --------------------------------------------------------------------
+|            ----------      -----------      --------               |
+|           | Gatherer | => | Processor | => | Sender |              |                          
+|            ----------      -----------      --------               |
+ --------------------------------------------------------------------
+```
+
+## Plugins
+
+Plugin is the minimal components in the module. There are 2 kinds of plugins in Satellite, which are sharing plugins and normal plugins.
```
wu-sheng, line 38: Please consider renaming the `kind`. Such as, Sateliite has 2 plugin catalogs

Makefile
```diff
@@ -70,3 +70,12 @@ verify: clean license lint test
...
 	-rm -rf coverage.txt
+
+.PHONY: build
+build: deps windows linux darwin
+
+
+.PHONY: $(PLATFORMS)
+$(PLATFORMS):
+	mkdir -p $(OUT_DIR)
+	GOOS=$(os) GOARCH=$(ARCH) $(GO_BUILD) $(GO_BUILD_FLAGS) -ldflags "$(GO_BUILD_LDFLAGS)" -o $(OUT_DIR)/$(BINARY)-$(VERSION)-$(os)-$(ARCH) ./cmd
```
gxthrj, line 81: No newline at end of file.

cmd/command.go
```diff
@@ -0,0 +1,51 @@
...
+
+var (
+	cmdStart = cli.Command{
+		Name:  "start",
+		Usage: "start satellite",
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:    "config, c",
+				Usage:   "Load configuration from `FILE`",
+				EnvVars: []string{"MOSN_CONFIG"},
```
gxthrj, line 35: mosn config?

EvanLjp: sorry ,I learned the use of golang CLI through MOSN , already fix it

docs/design/plugin_structure.md
```diff
@@ -0,0 +1,44 @@
...
+
+1. Find the plugin name in the input config according to the fixed key `plugin_name`.
+2. Find plugin type according to the plugin category(interface) and the plugin name.
+3. Create an empty plugin.
+4. Initialize the plugin according to the merged config, which is created by the input config and the default config.
+
+
+
+## Plugin usage in Satellite
+Nowadays, the numbers of [sharing Plugin](module_design.md) is 2, and the numbers of the other [normal Plugin](module_design.md) is 7.
```
wu-sheng, line 31: Don't put the number in the doc, it will change from time to time.

internal/pkg/log/log.go
```diff
@@ -0,0 +1,146 @@
...
+	"strings"
+	"sync"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Default logger config.
+const (
+	defaultLogPattern  = "%time [%level][%field] - %msg"
+	defaultTimePattern = "2006-01-02 15:04:05.001"
```
wu-sheng, line 32: Is this a pattern?

wu-sheng:
The new structure seems fine, but the `lock` is still there.

To all reviewers, I am going to OOO in the next 4 days, so, would not show up for reviewing details.

docs/design/plugin_structure.md
EvanLjp: okay

internal/pkg/log/log.go
EvanLjp: yes, it is the log pattern

EvanLjp:
> The new structure seems fine, but the `lock` is still there.
> 
> To all reviewers, I am going to OOO in the next 4 days, so, would not show up for reviewing details.

already remove the lock in the plugins, thx for your notice.

dist/licenses/LICENSE-cli
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
gxthrj, line 21: Need newline at end of file

dist/licenses/LICENSE-viper
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
gxthrj, line 21: ditto

surlymo (approved):
LGTM

gxthrj (commented):
LGTM

EvanLjp:
@gxthrj @wu-sheng @surlymo  I think the codes have achieved a three-approved goal.  Could u help me to approve it again, I would merge it?

wu-sheng (approved):

PR #11, opened 2020-12-28
Satellite Self-Observability

EvanLjp:
add prometheus telemetry


dist/licenses/LICENSE-prometheus-client-golang.txt
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
wu-sheng, line 1: Apache 2.0 license is not required to be copied. If NOTICE exists, you need to copy the content into `dist/NOTICE`.

dist/LICENSE
```diff
@@ -214,7 +214,7 @@ Apache 2.0 licenses
 
 The following components are provided under the Apache License. See project link for details.
 The text of each license is the standard Apache 2.0 license.
```
wu-sheng, line 216: Note this sentence.

configs/satellite_config.yaml
```diff
@@ -17,23 +17,30 @@
...
+  common_config:
+    pipe_name: sharing
   clients:
     - plugin_name: "grpc-client"
       k: v
   servers:
     - plugin_name: "grpc-server"
       k: v
-namespaces:
+pipes:
```
wu-sheng, line 37: What does `pipes` mean?

plugins/server/prometheus/Prometheus_test.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func TestServer_Start(t *testing.T) {
+	server, err := initPrometheusServer(make(plugin.Config))
+	if err != nil {
+		t.Fatalf("cannot init the prometheus server: %v", err)
+	}
+	_ = server.Start()
+	time.Sleep(time.Second)
+	response, err := http.Get("http://127.0.0.1" + server.Address + server.Endpoint)
```
wu-sheng, line 56: You have only tested the Prometheus endpoint is open, right?

internal/pkg/plugin/definition.go
```diff
@@ -33,12 +41,24 @@ type SharingPlugin interface {
...
-// NameField is a required field in Config.
-const NameField = "plugin_name"
+// GetPluginName returns a plugin name generated by the packages.
+func GetPluginName(p Plugin) string {
+	t := reflect.TypeOf(p)
+	if t.Kind() == reflect.Ptr {
+		t = t.Elem()
+	}
+	arr := strings.Split(t.PkgPath(), "/")
+	if len(arr) < 2 {
```
surlymo, line 60: magic num

internal/pkg/plugin/definition.go
```diff
@@ -33,12 +41,24 @@ type SharingPlugin interface {
...
 	Close() error
 }
 
 // Config is used to initialize the DefaultInitializingPlugin.
 type Config map[string]interface{}
 
-// NameField is a required field in Config.
-const NameField = "plugin_name"
+// GetPluginName returns a plugin name generated by the packages.
+func GetPluginName(p Plugin) string {
```
surlymo, line 54: why not just return the string name?

plugins/fallbacker/timer/timer_fallbacker.go
```diff
@@ -20,13 +20,15 @@ package timer
...
-	"github.com/apache/skywalking-satellite/internal/pkg/event"
+	"github.com/apache/skywalking-satellite/internal/pkg/config"
+	"github.com/apache/skywalking-satellite/internal/satellite/event"
 	"github.com/apache/skywalking-satellite/plugins/forwarder/api"
 )
 
 // Fallbacker is a timer fallbacker when forward fails. `latencyFactor` is the standard retry duration,
 // and the time for each retry is expanded by 2 times until the number of retries reaches the maximum.
 type Fallbacker struct {
+	config.CommonFields
```
surlymo, line 31: why all need this struct

internal/pkg/plugin/registry.go
```diff
@@ -100,4 +102,16 @@ func initializing(plugin Plugin, cfg Config) {
 	if err := v.Unmarshal(plugin); err != nil {
 		panic(fmt.Errorf("cannot inject  the config to the %s plugin, the error is %v", plugin.Name(), err))
 	}
+	cf := reflect.ValueOf(plugin).Elem().FieldByName(config.CommonFieldsName)
+	if !cf.IsValid() {
+		panic("%s plugin must have a field named CommonField.")
```
kezhenxu94, line 107:
```suggestion
		panic(fmt.Errorf("%s plugin must have a field named CommonField.", plugin.Name()))
```

internal/pkg/plugin/registry.go
```diff
@@ -47,12 +49,12 @@ func RegisterPlugin(plugin Plugin) {
 	for pCategory, pReg := range reg {
 		if v.Type().Implements(pCategory) {
 			pReg[plugin.Name()] = v
-			fmt.Printf("register %s %s successfully ", plugin.Name(), v.Type().String())
+			fmt.Printf("register %s %s successfully\n", plugin.Name(), v.Type().String())
```
kezhenxu94, line 52: I didn't notice this before, but `fmt.Printf` doesn't allow users to specify verbosity of logs, many other places are using `fmt.Printf` too

internal/pkg/plugin/registry.go
```diff
@@ -100,4 +102,16 @@ func initializing(plugin Plugin, cfg Config) {
 	if err := v.Unmarshal(plugin); err != nil {
 		panic(fmt.Errorf("cannot inject  the config to the %s plugin, the error is %v", plugin.Name(), err))
 	}
+	cf := reflect.ValueOf(plugin).Elem().FieldByName(config.CommonFieldsName)
+	if !cf.IsValid() {
+		panic("%s plugin must have a field named CommonField.")
+	}
+	for i := 0; i < cf.NumField(); i++ {
+		tagVal := cf.Type().Field(i).Tag.Get("mapstructure")
```
kezhenxu94, line 110: Use constant `TagName`

tools/install_protoc.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+
+if [[ "$(uname)"=="Darwin" || "$(expr substr $(uname -s) 1 10)"=="MINGW32_NT" ]];then
```
kezhenxu94, line 21: I saw there is a binary for MacOS too, why not support MacOS here as well?

wu-sheng: I think he is setting CI? Which only include Ubuntu for now? We should set up the MacOS for CI testing too.

internal/satellite/telemetry/telemetry.go
```diff
@@ -0,0 +1,53 @@
...
+	registry   *prometheus.Registry
+	Registerer prometheus.Registerer // The register is for adding metrics to the registry.
+	Gatherer   prometheus.Gatherer   // The gatherer is for fetching metrics from the registry.
+)
+
+// Config defines the common telemetry labels.
+type Config struct {
+	Service   string `mapstructure:"service"`   // The service name.
+	Namespace string `mapstructure:"namespace"` // The kubernetes namespace.
+	Pod       string `mapstructure:"pod"`       // The kubernetes pod name.
```
hanahmily, line 35: IIRC, we should not introduce Kubernetes notions into satellite since it's designed to deployed on VM as well. 

How about using more general and SkyWalking-style concepts, for example, `Cluster`, `Service` and `Instance`

plugins/server/prometheus/Prometheus.go
```diff
@@ -0,0 +1,83 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
```
hanahmily, line 1: Why the file name is a capical letter?

plugins/server/prometheus/Prometheus_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"testing"
+	"time"
+
+	"github.com/apache/skywalking-satellite/internal/pkg/log"
+	"github.com/apache/skywalking-satellite/internal/pkg/plugin"
+	"github.com/apache/skywalking-satellite/internal/satellite/telemetry"
+	"github.com/apache/skywalking-satellite/plugins/server/api"
+)
+
+func initPrometheusServer(cfg plugin.Config) (*Server, error) {
```
hanahmily, line 33: Could you use different endpoints and ports instead of defaults.

.github/workflows/build-and-test.yaml
```diff
@@ -36,12 +36,6 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
 
-      - name: Install Protoc
-        uses: arduino/setup-protoc@v1
```
hanahmily, line 40: @kezhenxu94 Could we use a third part plugin now?

tools/install_protoc.sh
hanahmily: In this PR, he's using a third-party plugin instead of this bash to install protoc

hanahmily (commented):
Some nits

.github/workflows/build-and-test.yaml
kezhenxu94: > @kezhenxu94 Could we use a third part plugin now?

This kind of actions are blocked, unless we get enough reasons to persuade the infra team to unblock it.

tools/install_protoc.sh
kezhenxu94: > I think he is setting CI? Which only include Ubuntu for now? We should set up the MacOS for CI testing too.

Locally, when you run `make deps` to setup the dev env, this is executed and if you didn't install protoc correctly, it fails.

wu-sheng:
> In this PR, he's using a third-party plugin instead of this bash to install protoc

I think 3rd party github action has been blocked by INFRA. @hanahmily

internal/satellite/module/gatherer/receiver_gatherer.go
```diff
@@ -41,51 +44,75 @@ type ReceiverGatherer struct {
...
-	log.Logger.Infof("receiver gatherer module of %s namespace is preparing", r.config.NamespaceName)
-	r.runningReceiver.RegisterHandler(r.runningServer)
+	log.Logger.Infof("receiver gatherer module of %s namespace is preparing", r.config.PipeName)
+	r.runningReceiver.RegisterHandler(r.runningServer.GetServer())
 	if err := r.runningQueue.Initialize(); err != nil {
-		log.Logger.Infof("the %s queue of %s namespace was failed to initialize", r.runningQueue.Name(), r.config.NamespaceName)
+		log.Logger.Infof("the %s queue of %s namespace was failed to initialize", r.runningQueue.Name(), r.config.PipeName)
 		return err
 	}
+	r.receiveCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
```
gxthrj, line 59: I am not very clear about this. Should the metrics here be defined in plugins ? Or we supports prometheus in `Gather` by default?

configs/satellite_config.yaml
EvanLjp: the namespace concept is similar to the Kubernetes namespaces, so I change the original namespaces concept to the new name call pipe. Each pipe has one gatherer, one processor, and one sender.

plugins/server/prometheus/Prometheus_test.go
EvanLjp: yes

internal/pkg/plugin/definition.go
EvanLjp: Use the directory structure to ensure that the plug-in name is unique

plugins/fallbacker/timer/timer_fallbacker.go
EvanLjp: the common field is inject by boot process.
for example:  a plugins want to know which pipe belong to, the commonField is for this.

internal/satellite/telemetry/telemetry.go
EvanLjp: no problem

internal/satellite/module/gatherer/receiver_gatherer.go
EvanLjp: the feature is to monitor self status.  Could you explain that again? I'm not very clear

plugins/server/prometheus/Prometheus_test.go
wu-sheng: Got it.

configs/satellite_config.yaml
wu-sheng: Do you update the doc? I think the `namespace` has been mentioned there.

EvanLjp: sorry, thx for your notice

plugins/server/prometheus/Prometheus_test.go
EvanLjp: image
could set custom values

EvanLjp: image

.github/workflows/build-and-test.yaml
EvanLjp: image
@hanahmily  i use shell to install protoc

EvanLjp:
@gxthrj @nic-chen  @kezhenxu94  how about to use init() in logger? can't use init() in Logger brings up a lot of chore things.

internal/satellite/module/gatherer/receiver_gatherer.go
gxthrj: > the feature is to monitor self status

OK, This is the metric for satellite self.
I mistakenly thought that this is an extension of Prometheus Gather.

wu-sheng: @EvanLjp I think you should rename this PR to `Satellite Self-Observability` :) I used to think this is for scratch from prometheus endpoint.

tools/install_protoc.sh
wu-sheng: I tried, I have to install the protoc manually.

wu-sheng (approved):
Generally good to me.

gxthrj (approved):
LGTM

surlymo (approved):
LGTM

PR #12, opened 2021-01-01
gRPC server & native log receiver

EvanLjp:
The gRPC server & log receiver is for receiving the logData in protocol v3.


plugins/receiver/grpclog/receiver.go
```diff
@@ -0,0 +1,71 @@
...
+	outputChannel chan *protocol.Event // The output channel of grpc log
+
+}
+
+func (r *Receiver) Name() string {
+	return plugin.GetPluginName(r)
+}
+
+func (r *Receiver) Description() string {
+	return "this is a grpc log receiver"
```
wu-sheng, line 48:
```suggestion
	return "This is a receiver for SkyWalking native logging format, which is defined at https://github.com/apache/skywalking-data-collect-protocol/blob/master/logging/Logging.proto."
```

plugins/receiver/grpclog/receiver.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/apache/skywalking-satellite/protocol/gen-codes/satellite/protocol"
+)
+
+type Receiver struct {
+	config.CommonFields
+	MaxBufferSpaces int `mapstructure:"max_buffer_spaces"` // The max buffer events.
+
+	server        *grpc.Server
+	service       *LogReportService
+	outputChannel chan *protocol.Event // The output channel of grpc log
```
wu-sheng, line 39: Is this a channel for gRPC server to handle the network request? If so, the comments seem misleading.

plugins/receiver/grpclog/receive_service.go
```diff
@@ -0,0 +1,59 @@
...
+
+const eventName = "grpc-log-event"
+
+type LogReportService struct {
+	receiveChannel chan *protocol.Event
+	logging.UnimplementedLogReportServiceServer
+}
+
+func (s *LogReportService) Collect(stream logging.LogReportService_CollectServer) error {
+	startTime := time.Now()
```
nic-chen, line 38: how about `startTime := time.Now().Unix()`

plugins/receiver/grpclog/receive_service.go
```diff
@@ -0,0 +1,58 @@
...
+		logData, err := stream.Recv()
+		if err == io.EOF {
+			return stream.SendAndClose(&common.Commands{})
+		}
+		if err != nil {
+			return err
+		}
+		e := &protocol.Event{
+			Name:      eventName,
+			Timestamp: time.Now().Unix(),
```
wu-sheng, line 48: I think all SkyWalking timestamp should be in millisecond. `Unix` is `number of seconds`. You need to convert nanosecond.

plugins/receiver/grpclog/receive_service.go
```diff
@@ -0,0 +1,58 @@
...
+
+	common "skywalking/network/common/v3"
+	logging "skywalking/network/logging/v3"
+
+	"github.com/apache/skywalking-satellite/protocol/gen-codes/satellite/protocol"
+)
+
+const eventName = "grpc-log-event"
+
+type LogReportService struct {
```
wu-sheng, line 32: Could you rename the `receive_service.go` according to this? Same request for receiver.go. By the rule you are using now, we will face duplicated names in all plugins.

EvanLjp: no problem

wu-sheng (approved):
LGTM

plugins/server/grpc/server.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+type Server struct {
+	config.CommonFields
+	Address              string `mapstructure:"address"`                // The address of grpc server.
+	Network              string `mapstructure:"network"`                // The network of grpc.
+	MaxRecvMsgSize       int    `mapstructure:"max_recv_msg_size"`      // The max size of the received log.
+	MaxConcurrentStreams uint32 `mapstructure:"max_concurrent_streams"` // The max concurrent stream channels.
+	TLSCertFile          string `mapstructure:"tls_cert_file"`          // The TLS cert file path.
+	TLSKeytFile          string `mapstructure:"tls_key_file"`           // The TLS key file path.
```
kezhenxu94, line 38: s/TLSKeytFile/TLSKeyFile

kezhenxu94 (commented):
Just one nit, others LGTM

plugins/receiver/nativelog/receiver.go
```diff
@@ -0,0 +1,72 @@
...
+
+	logging "skywalking/network/logging/v3"
+
+	"github.com/apache/skywalking-satellite/internal/pkg/config"
+	"github.com/apache/skywalking-satellite/internal/pkg/plugin"
+	"github.com/apache/skywalking-satellite/protocol/gen-codes/satellite/protocol"
+)
+
+type Receiver struct {
+	config.CommonFields
```
surlymo, line 34: trace\metrics\log's grpc receiver looks like almost the same with each other? why not use the same grpc protocol/receiver/report_service to unify the model

EvanLjp: the protocol is not the same on the different scopes. https://github.com/apache/skywalking-data-collect-protocol

wu-sheng: @EvanLjp I think @surlymo 's idea is, more settings could be moved into a new parent, such as a thing called GRPCReceiver, to reduce the duplicated fields like `server`, `MaxBufferSpaces`,

EvanLjp: there is only one config named MaxBufferSpaces in grpclog-receiver. The config is for buffer some data before enqueueing. I think different receivers may have different priorities, so the receiver can have different configurations. The other configurations, such that the server, is a reference.

I think @surlymo 's idea is to use one collect service to work in multi-data scopes.

wu-sheng: > I think @surlymo 's idea is to use one collect service to work in multi-data scopes.But this may not fit the plug-in concept very well

This is not working. True.

____
I want to be back a little about my idea. 
```go
type Receiver struct {
	config.CommonFields
	MaxBufferSpaces int `mapstructure:"max_buffer_spaces"` // The max buffer events.

	server        *grpc.Server
	service       *LogReportService    // The gRPC request handler for logData.
	outputChannel chan *protocol.Event // The channel is to bridge the LogReportService and the Gatherer to delivery the data.

}
```
I mean is, this struct is in `nativelog`, but actually, all receiver plugin will have similar configuration besides the bind service, right? I can't see how does this related to priority.

EvanLjp: I would polish it now

plugins/receiver/grpc/common.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"google.golang.org/grpc"
+
+	"github.com/apache/skywalking-satellite/protocol/gen-codes/satellite/protocol"
+)
+
+type CommonGRPCReceiverFields struct {
+	Server        *grpc.Server
+	OutputChannel chan *protocol.Event // The channel is to bridge the LogReportService and the Gatherer to delivery the data.
```
wu-sheng, line 28: As you moved the `OutputChannel` and `Server` into the CommonGRPCReceiverFields, should you consider to move the initialization into here too?

internal/pkg/plugin/definition.go
```diff
@@ -53,14 +55,16 @@ type Config map[string]interface{}
...
 	}
-	arr := strings.Split(t.PkgPath(), "/")
-	if len(arr) < nameParts {
-		return ""
+	path := t.PkgPath()
+	path = path[strings.LastIndex(path, PluginDir)+len(PluginDir)+1:]
+	arr := strings.Split(path, "/")
+	if len(arr) > 1 {
+		category := arr[0]
+		return strings.Join(arr[1:], "-") + "-" + category
```
gxthrj, line 67: The first feeling is so complicated.
`plugin` is an interface, why not organize name in the specific plugin implementation. And then we can use `plugin.Name()` here instead of L62~67.

plugins/receiver/grpc/nativelog/receiver.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type Receiver struct {
+	config.CommonFields
+	grpcreceiver.CommonGRPCReceiverFields
+	service *LogReportService // The gRPC request handler for logData.
+}
+
+func (r *Receiver) Name() string {
+	return plugin.GetPluginName(r)
```
gxthrj, line 41: Ditto.
I think `GetPluginName` is not necessary.  And this forces us to use reflection on the interface to get the implement name.
We can make rules that  setting `Name` in each plugin instance directly.

internal/pkg/plugin/definition.go
EvanLjp: The directory structure is an effective way to avoid duplication of plug-in names, and we recommend this util, but it is not mandatory

EvanLjp: @wu-sheng @surlymo  what do u think about it ?

wu-sheng: As you asked, personally, prefer to use `plugin.Name()` to simplify.

wu-sheng (approved):
LGTM. @EvanLjp Do you need anyone else to recheck?

EvanLjp:
> LGTM. @EvanLjp Do you need anyone else to recheck?

I think it is enough to merge.

PR #13, opened 2021-01-02
add kafka forwarder

EvanLjp:
add the Kafka log forwarder, the parts should be tested by e22, which would be added when e2e is built in Satellie.  Has passed the manual test, if possible, please merge first.@kezhenxu94 What do you think of Satellite E2E?


wu-sheng:
About the e2e, we need @arugal 's OAP implementation ready. And for your case, we need log-kafka implementation.

wu-sheng:
Also, you need the agent plugin implementation. Are you going to submit it? How do you plan to do that?

dist/licenses/LICENSE-kafka.txt
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2013 Shopify
```
kezhenxu94, line 1: The file name should be `LICENSE-sarama.txt`

kezhenxu94:
> @kezhenxu94 What do you think of Satellite E2E? 

I can help to set up the E2E here after OAP implementation is provided.

Humbertzhang:
I am interest in satellite, I can help to set up the E2E

wu-sheng:
The CIs fail, please recheck.

wu-sheng (approved):
Generally LGTM.

surlymo (approved):
LGTM

PR #16, opened 2021-01-17
feat: add http logging receiver

gxthrj:
Add a http receiver plugin .

PR #10, opened 2020-12-16
mmap-queue-plugin

EvanLjp:
the mmap plugin on design doc. The design doc and benchmark test report is on skywalking-satellite/docs/plugins/queue/mmap. Please let me know if you have any suggestions on the swap segments algorithm.

PR #18, opened 2021-01-20
Include go.sum|go.mod in CI consistency check

kezhenxu94:
go.sum|go.mod need to be checked too @EvanLjp , see https://github.com/apache/skywalking-satellite/pull/16#discussion_r560815717

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -48,7 +48,11 @@ jobs:
         run: make deps
 
       - name: Check generate
-        run: make gen && make build && make check
+        run: make gen && make build
+
+      - name: Check CI Consistency
+        if: "${{ matrix.go-version }}" == "1.15"
+        run: make check
 
       - name: Lint
         run: make lint
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -51,7 +51,7 @@ jobs:
         run: make gen && make build
 
       - name: Check CI Consistency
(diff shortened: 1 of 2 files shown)


EvanLjp:
But in the history cases, the `go mod tidy` is not stable in the different machines to add `// indirect`

kezhenxu94:
According to [the doc](https://github.com/golang/go/wiki/Modules#why-does-go-mod-tidy-record-indirect-and-test-dependencies-in-my-gomod), `go mod tidy` should be consistent in all platforms:

> go mod tidy also ensures your current go.mod reflects the dependency requirements for all possible combinations of OS, architecture, and build tags (as described here). In contrast, other commands like go build and go test only update go.mod to provide the packages imported by the requested packages under the current GOOS, GOARCH, and build tags (which is one reason go mod tidy might add requirements that were not added by go build or similar).

I checked your issue in #10 , which is caused by the `packr` tool because it used `GO111MODULE=on` so that it modifies the `go.{mod,sum}` always when no `packr2` is installed (this happens on GitHub Action, but not on your local machine unless you remove `$(go env GOPATH)/bin/packr2`), we don't use it and thus I remove it here.

Another problem is that you had `$(GO) mod tidy &> /dev/null`, causing the `go mod tidy` to run at background and the following command `git status -s` will detect nothing because `go mod tidy` has not finished yet (mostly because it cause some time), see [here](https://github.com/apache/skywalking-satellite/runs/1740240310) for the problem that was never detected before because of this .

EvanLjp (approved):

PR #19, opened 2021-01-23
polish plugins for e2e tests

EvanLjp:
polish plugins for e2e test

### DIFF
--- a/configs/satellite_config.yaml
+++ b/configs/satellite_config.yaml
@@ -21,7 +21,7 @@ logger:
   level: "info"
 
 telemetry:
-  cluster: namespaces
+  cluster: cluster1
   service: service1
   instance: instance1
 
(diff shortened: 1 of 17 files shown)


configs/satellite_config.yaml
```diff
@@ -21,41 +21,42 @@ logger:
   level: "info"
 
 telemetry:
-  namespace: namespaces
+  cluster: namespaces
```
wu-sheng, line 24: What is this renaming for? I also still see `namespace1` in the pipe_name. Could you be clear more?

EvanLjp: they are different concepts, one is the cluster concept, such that `Namespace` in the Kubernetes. Another one is the internal concept, which means running in the different pipes. Please ignore it now, I would add the env override mod later, which is similar withe the SkyWalking.

wu-sheng (approved):
LGTM

PR #22, opened 2021-01-30
add memory queue

EvanLjp:
add memory queue, please pay attention to the concurrency problem 
https://github.com/apache/skywalking/issues?q=is%3Aopen+mentions%3A%40me

### DIFF
--- a/plugins/queue/memory/queue.go
+++ b/plugins/queue/memory/queue.go
@@ -18,10 +18,10 @@
 package memory
 
 import (
-	"fmt"
-	"sync/atomic"
+	"github.com/enriquebris/goconcurrentqueue"
 
 	"github.com/apache/skywalking-satellite/internal/pkg/config"
+	"github.com/apache/skywalking-satellite/internal/pkg/log"
 	"github.com/apache/skywalking-satellite/internal/satellite/event"
 	"github.com/apache/skywalking-satellite/plugins/queue/api"
 	"github.com/apache/skywalking-satellite/protocol/gen-codes/satellite/protocol"
@@ -29,23 +29,15 @@ import (
 
 const (
 	Name = "memory-queue"
-	// discard strategy
-	discardOldest = "DISCARD_OLDEST"
-	discardLatest = "DISCARD_LATEST"
 )
 
 type Queue struct {
 	config.CommonFields
 	// config
-	EventBufferSize int64  `mapstructure:"event_buffer_size"` // The maximum buffer event size.
-	DiscardStrategy string `mapstructure:"discard_strategy"`  // The discard strategy.
+	EventBufferSize int `mapstructure:"event_buffer_size"` // The maximum buffer event size.
 
 	// components
-	queue []*protocol.Event
-	// The position continuously increasing, but don't worry, it can run for another 1067519911 days at 10W OPS.
-	readPos  int64
-	writePos int64
-	count    int64
+	buffer *goconcurrentqueue.FixedFIFO
 }
 
 func (q *Queue) Name() string {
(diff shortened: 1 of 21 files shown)


plugins/queue/memory/queue.go
```diff
@@ -0,0 +1,121 @@
...
+	return "this is a memory queue to buffer the input event."
+}
+
+func (q *Queue) DefaultConfig() string {
+	return `
+# The maximum buffer event size.
+event_buffer_size: 5000
+# The discard strategy when facing the full condition.
+# There are 2 strategies, which are LOST_THE_OLDEST_ONE and LOST_THE_NEW_ONE. 
+discard_strategy: LOST_THE_OLDEST_ONE
```
kezhenxu94, line 65: Since this is called "discard strategy", why not name them as `discardOldest` and `discardLatest`, or even only use `latest` and `oldest` since you have the context here as the key

EvanLjp: thx

EvanLjp: fixed

plugins/queue/memory/queue.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+	} else {
+		atomic.AddInt64(&q.count, 1)
+	}
+	q.queue[q.writePos%q.count] = e
+	q.writePos++
+	return nil
+}
+
+func (q *Queue) Pop() (*api.SequenceEvent, error) {
```
kezhenxu94, line 96: The name `Push` and `Pop` are more about Stacks, for queues, it's more intuitive to say `enqueue` and `dequeue`

EvanLjp: fixed

plugins/queue/memory/queue.go
```diff
@@ -0,0 +1,121 @@
...
+func (q *Queue) Enqueue(e *protocol.Event) error {
+	if q.isFull() {
+		switch q.DiscardStrategy {
+		case discardLatest:
+			return api.ErrFull
+		case discardOldest:
+			atomic.AddInt64(&q.readPos, 1)
+		}
+	} else {
+		atomic.AddInt64(&q.count, 1)
```
wu-sheng, line 89: This add could be running in concurrency, then the count could be greater `EventBufferSize`, then  `q.writePos%q.count` could be `out of bound`

wu-sheng: If you want to lower the chance of index, and keep the discard rule as expected, my understanding, you should make the all index manipulation in a lock somehow.

plugins/queue/memory/queue.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (q *Queue) Close() error {
+	return nil
+}
+
+func (q *Queue) Ack(_ event.Offset) {
+}
+
+func (q *Queue) isEmpty() bool {
+	return q.count == 0
```
nic-chen, line 116: should we use atomic load here？

plugins/queue/memory/queue.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (q *Queue) isFull() bool {
+	return q.count == q.EventBufferSize
```
nic-chen, line 120: ditto

plugins/queue/memory/queue_test.go
```diff
@@ -0,0 +1,120 @@
...
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			q, err := initMemoryQueue(tt.args)
+			if err != nil {
+				t.Fatalf("cannot init the memory queue: %v", err)
+			}
+			for i := 0; i < num; i++ {
```
nic-chen, line 80: we should have test case about exceed the buffer size

plugins/queue/mmap/branchmark_test.go
```diff
@@ -50,13 +50,13 @@ func cleanBenchmarkQueue(b *testing.B, q api.Queue) {
 	}
 }
```
gxthrj, line 52: Should the file name be `benchmark_test.go`?

nic-chen:
ping @EvanLjp

EvanLjp:
Sorry,recently，i am quite busy on my work，i would force on satellite after tomorrow.

EvanLjp:
@nic-chen @wu-sheng I add a thread safe lib called goconcurrentqueue to solve the problem. I will supplement the code document recently for preparation before release.

plugins/queue/memory/queue.go
```diff
@@ -0,0 +1,87 @@
...
+	Name = "memory-queue"
+)
+
+type Queue struct {
+	config.CommonFields
+	// config
+	EventBufferSize int `mapstructure:"event_buffer_size"` // The maximum buffer event size.
+
+	// components
+	buffer *goconcurrentqueue.FixedFIFO
```
wu-sheng, line 40: Forgive me don't read their documentation first, is this FIFO queue being blocked when the queue is full? If so, do they provide any policy to abandon data or something?

EvanLjp: yes, the queue would reject the new data.

wu-sheng (approved):
LGTM, we need a lot of polish about the document before our first release.

EvanLjp:
> LGTM, we need a lot of polish about the document before our first release.

I will continue to do this from today

wu-sheng:
@gxthrj @nic-chen Please confirm this PR.

nic-chen (approved):
LGTM

PR #24, opened 2021-02-10
Unify the build process on all platforms

kezhenxu94:
Because some plugins are not available on Windows, those plugins are separated in individual package and are only built on available platforms, using `+build` flag.


EvanLjp (approved):
Thx for your helping.

PR #26, opened 2021-02-13
add start up script

EvanLjp:
add start-up script for start satellite. But I am not familiar with the BAT script and don't have an environment for testing, who can help to write the BAT startup script?

If no one familiar with it, I would prepare an environment and learn the bat script to finish it later.


configs/satellite_config.yaml
```diff
@@ -29,8 +29,6 @@ telemetry:
 
 # The sharing plugins referenced by the specific plugins in the different pipes.
 sharing:
-  common_config:
```
nic-chen, line 32: why should we remove these two lines ?

dist/bin/startup.sh
```diff
@@ -0,0 +1,47 @@
...
+
+eval exec "$START_UP_PROCESS" start --config="$CONFIG_DIR"/satellite_config.yaml 1> "$LOG_FILE_LOCATION" 2>&1 &
+
+if [ $? -eq 0 ]; then
+  sleep 1
+	echo "SkyWalking Satellite started successfully!"
+else
+	echo "SkyWalking Satellite started failure!"
+	exit 1
+fi
```
nic-chen, line 47: need a blank line at the end..

dist/bin/startup.sh
```diff
@@ -0,0 +1,47 @@
...
+if [ ! -d "${LOG_DIR}" ]; then
+    mkdir -p "${LOG_DIR}"
+fi
+
+if uname -s | grep Darwin; then
+  START_UP_PROCESS=$(find "$BIN_DIR" -name "skywalking-satellite*darwin*")
+elif uname -s | grep Linux; then
+  START_UP_PROCESS=$(find "$BIN_DIR" -name "skywalking-satellite*linux*")
+else
+  echo "Sorry, the official startup process does not support your system."
```
nic-chen, line 36: should we exit here?

configs/satellite_config.yaml
EvanLjp: already add default config in codes

dist/bin/startup.sh
EvanLjp: fixed

wu-sheng (commented):
Generally good, but document is not updated.

dist/bin/startup.sh
```diff
@@ -0,0 +1,48 @@
...
```
kezhenxu94, line 36: You can use `START_UP_PROCESS=$(find "$BIN_DIR" -name "skywalking-satellite*windows*")` here, Windows users can still use this shell script if they use something like Git Bash or Cygwin, otherwise this line of code will never be reached IMO, because Powershell or Command Prompt won't execute `.sh` file at all

dist/bin/startup.sh
```diff
@@ -0,0 +1,48 @@
...
+  exit 1
+fi
+
+eval exec "$START_UP_PROCESS" start --config="$CONFIG_DIR"/satellite_config.yaml 1> "$LOG_FILE_LOCATION" 2>&1 &
+
+if [ $? -eq 0 ]; then
+  sleep 1
+	echo "SkyWalking Satellite started successfully!"
```
kezhenxu94, line 44:
```suggestion
  echo "SkyWalking Satellite started successfully!"
```

dist/bin/startup.sh
```diff
@@ -0,0 +1,48 @@
...
+else
+	echo "SkyWalking Satellite started failure!"
+	exit 1
```
kezhenxu94, line 47:
```suggestion
  echo "SkyWalking Satellite started failure!"
  exit 1
```

kezhenxu94 (commented):
> If no one familiar with it, I would prepare an environment and learn the bat script to finish it later.

If you don't have a Windows system at hand, I think it's OK to leave it to the community, but make sure to complete the documentation so that Windows users can start it up even if they don't have a convenient script.

EvanLjp:
> Generally good, but document is not updated.

already update it,  Is there anything else I should pay attention to?
image

EvanLjp:
Do you have any suggestions about the documentation? If there is no problem, I will do a final test with disk queue or memory queue recently and want to release it this week? @wu-sheng @kezhenxu94

kezhenxu94:
> Do you have any suggestions about the documentation? If there is no problem, I will do a final test with disk queue or memory queue recently and want to release it this week? @wu-sheng @kezhenxu94

Another thing I want to remind is that, as this is typically deployed as a sidecar, we'd better provide a Docker image for convenience (so a Dockerfile is also needed for releasing), before the 1st release

EvanLjp:
ok，would add it later > > Do you have any suggestions about the documentation? If there is no problem, I will do a final test with disk queue or memory queue recently and want to release it this week? @wu-sheng @kezhenxu94
> 
> 
> 
> Another thing I want to remind is that, as this is typically deployed as a sidecar, we'd better provide a Docker image for convenience (so a Dockerfile is also needed for releasing), before the 1st release

docs/en/setup/README.md
```diff
@@ -2,10 +2,9 @@
 First and most important thing is, SkyWalking Satellite startup behaviours are driven by configs/satellite_config.yaml. Understanding the setting file will help you to read this document.
 
 ## Startup script
-The startup script is /bin/skywalking-satellite-{version}-{plateform}-amd64. 
-1. Start SkyWalking Satellite.
+Startup Script
 ```shell script
-./bin/skywalking-satellite-{version}-{plateform}-amd64 start --config=./configs/satellite_config.yaml
+/bin/startup.sh
```
gxthrj, line 7: Should be a relative directory

EvanLjp:
@kezhenxu94  thx for review, already add the satellite docker build script.

docker/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    tar -xzf "$DIST_NAME.tgz"; \
+    rm -rf "$DIST_NAME.tgz"; \
+    mkdir /skywalking-satellite; \
+    mv "$DIST_NAME"/* /skywalking-satellite;
+
+WORKDIR /skywalking-satellite
+
+EXPOSE 12800 11800 1234
+
+ENTRYPOINT ["/bin/sh","bin/startup.sh"]
```
kezhenxu94, line 33: - Do not base on `bash` image, you can save 80% (21MB vs 110MB) if you base on `scratch`.
- Do not use `bin/startup.sh`, you need this on non-containerized machine because you need to determine the os / platform, when in docker environment, you can determine which platform / os it is (because you choose it)

EvanLjp: 1. finally, use busybox because of sh
2. add a new entrypoint.sh

gxthrj (approved):
LGTM

nic-chen (approved):

PR #27, opened 2021-02-19
polish codes for release & add performance test doc

EvanLjp:
This is maybe the last PR for 0.1.0 milestone. Thx for reviewing it.


Makefile
```diff
@@ -73,11 +75,16 @@ clean: tools
 	-rm -rf coverage.txt
 
 .PHONY: build
-build: clean deps linux darwin windows
+build: clean deps
+	rm -rf bin/*
```
kezhenxu94, line 79: This line should be in `clean` goal?

wu-sheng:
> This is maybe the last PR for 0.0.1 milestone. Thx for reviewing it.

Version incorrect. Your release is 0.1.0. The last version number is for bug hot-fix(emergency) only in SkyWalking

Makefile
EvanLjp: In the local test ， I build some different version docker images and push them to a private repository. Because of using the find command in the start script， the docker started failed when having history file. If needed to the original version, I would rollback it.

kezhenxu94: I polish the docker image, please review https://github.com/apache/skywalking-satellite/pull/28 and rebase this on that after merging it

docs/en/setup/plugins/queue_mmap-queue.md
```diff
@@ -4,17 +4,15 @@ This is a memory mapped queue to provide the persistent storage for the input ev
 ## DefaultConfig
 ```yaml
 # The size of each segment. Default value is 128K. The unit is Byte.
-segment_size: 131072
+segment_size: 262114
```
kezhenxu94, line 7: do you mean `262144`? `262144 == 256K` while `262114` is odd. Also need to update the comment above (line 6)

plugins/queue/mmap/queue.go
```diff
@@ -87,17 +87,15 @@ func (q *Queue) Description() string {
 func (q *Queue) DefaultConfig() string {
 	return `
 # The size of each segment. Default value is 128K. The unit is Byte.
-segment_size: 131072
+segment_size: 262114
```
kezhenxu94, line 90: Ditto

kezhenxu94 (dismissed):
Basically LGTM, just 2 nits

gxthrj (approved):

nic-chen (commented):
should we add some e2e test cases before release ?

EvanLjp:
> should we add some e2e test cases before release ?

@kezhenxu94 Do we have some ways to do it?

kezhenxu94:
> > should we add some e2e test cases before release ?
> 
> @kezhenxu94 Do we have some ways to do it?

The added docker-compose.yaml in #28 can be used to do E2E tests, @Humbertzhang @fgksgf do you have time to try to set it up in this repo?

kezhenxu94 (approved):

PR #28, opened 2021-02-20
Optimize Docker image and startup script

kezhenxu94:
Reduce the docker image size, building process, and start script.

The image size is reduced from ~40MB to ~15MB


wu-sheng (approved):

EvanLjp (approved):

EvanLjp:
LGTM

PR #30, opened 2021-02-20
add timer fallbacker unit test

EvanLjp:
add timer fallbacker unit test

### DIFF
--- a/plugins/fallbacker/timer/timer_fallbacker.go
+++ b/plugins/fallbacker/timer/timer_fallbacker.go
@@ -25,16 +25,18 @@ import (
 	"github.com/apache/skywalking-satellite/plugins/forwarder/api"
 )
 
-// Fallbacker is a timer fallbacker when forward fails. `latencyFactor` is the standard retry duration,
-// and the time for each retry is expanded by 2 times until the number of retries reaches the maximum.
+const Name = "timer-fallbacker"
+
+// Fallbacker is a timer fallbacker when forward fails.
 type Fallbacker struct {
 	config.CommonFields
-	maxTimes      int `mapstructure:"max_times"`
-	latencyFactor int `mapstructure:"latency_factor"`
+	MaxTimes       int `mapstructure:"max_times"`
+	LatencyFactor  int `mapstructure:"latency_factor"`
+	MaxLatencyTIme int `mapstructure:"max_latency_time"`
 }
 
 func (t *Fallbacker) Name() string {
-	return "timer-fallbacker"
+	return Name
 }
 
 func (t *Fallbacker) Description() string {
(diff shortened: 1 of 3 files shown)


wu-sheng (approved):

plugins/fallbacker/timer/timer_fallbacker.go
```diff
@@ -25,16 +25,18 @@ import (
...
-// and the time for each retry is expanded by 2 times until the number of retries reaches the maximum.
+const Name = "timer-fallbacker"
+
+// Fallbacker is a timer fallbacker when forward fails.
 type Fallbacker struct {
 	config.CommonFields
-	maxTimes      int `mapstructure:"max_times"`
-	latencyFactor int `mapstructure:"latency_factor"`
+	MaxTimes       int `mapstructure:"max_times"`
+	LatencyFactor  int `mapstructure:"latency_factor"`
```
kezhenxu94, line 34: s/LatencyFactor/ExponentialBackoff

plugins/fallbacker/timer/timer_fallbacker.go
```diff
@@ -25,16 +25,18 @@ import (
...
-// Fallbacker is a timer fallbacker when forward fails. `latencyFactor` is the standard retry duration,
-// and the time for each retry is expanded by 2 times until the number of retries reaches the maximum.
+const Name = "timer-fallbacker"
+
+// Fallbacker is a timer fallbacker when forward fails.
 type Fallbacker struct {
 	config.CommonFields
-	maxTimes      int `mapstructure:"max_times"`
-	latencyFactor int `mapstructure:"latency_factor"`
+	MaxTimes       int `mapstructure:"max_times"`
```
kezhenxu94, line 33: s/MaxTimes/MaxAttempts

plugins/fallbacker/timer/timer_fallbacker.go
```diff
@@ -43,17 +45,25 @@ func (t *Fallbacker) Description() string {
...
+# The max retry latency time.(Time unit is millisecond.)
+max_latency_time: 5000
 `
 }
 
 func (t *Fallbacker) FallBack(batch event.BatchEvents, forward api.ForwardFunc) bool {
-	currentLatency := t.latencyFactor
-	for i := 1; i < t.maxTimes; i++ {
+	currentLatency := t.LatencyFactor
+	for i := 1; i < t.MaxTimes; i++ {
```
kezhenxu94, line 60:
```suggestion
	for i := 1; i <= t.MaxTimes; i++ {
```

plugins/fallbacker/timer/timer_fallbacker.go
```diff
@@ -43,17 +45,25 @@ func (t *Fallbacker) Description() string {
...
 		time.Sleep(time.Duration(currentLatency) * time.Millisecond)
 		if err := forward(batch); err != nil {
 			currentLatency *= 2
+			if currentLatency > t.MaxLatencyTIme {
```
kezhenxu94, line 64:
```suggestion
			if currentLatency >= t.MaxLatencyTIme {
```

plugins/fallbacker/timer/timer_fallbacker.go
```diff
@@ -43,17 +45,25 @@ func (t *Fallbacker) Description() string {
...
+				currentLatency = t.MaxLatencyTIme
```
kezhenxu94, line 65: ??

```suggestion
				return false
```

kezhenxu94: Add this test case please

kezhenxu94 (commented):
Does the max attempts includes the original one (when the max attempts is set to 3 and it failed and comes to fallbacker, you only retry 3-1=2 times) ? If so ignore some of my comments. But others should be addressed.

kezhenxu94:
> Does the max attempts includes the original one (when the max attempts is set to 3 and it failed and comes to fallbacker, you only retry 3-1=2 times) ? If so ignore some of my comments. But others should be addressed.

plugins/fallbacker/timer/timer_fallbacker.go
EvanLjp: when trigger the fallbacker, already trigger once, so from 1 to add

PR #33, opened 2021-02-22
fix release src cannot compile bug

EvanLjp:
fix release src cannot compile bug

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ clean: tools
 	-rm -rf coverage.txt
 
 .PHONY: build
-build: clean deps gen-codes linux darwin windows
+build: clean gen-codes deps linux darwin windows
 
 .PHONY: check
 check: clean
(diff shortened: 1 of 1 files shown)


kezhenxu94:
Seems I can't still compile from the src.tgz made from this branch?

EvanLjp:
image
maybe could compile on my mac.

kezhenxu94:
@EvanLjp see my update 👆

wu-sheng:
@kezhenxu94 @EvanLjp I think this is why we use a `shell` rather than a simple make file to build src release.

@kezhenxu94 didn't do `git submodule init/update`, but run `make release-src`, then still src is not complete.

wu-sheng:
https://github.com/apache/skywalking/blob/master/tools/releasing/create_source_release.sh Like I asked, take a look at the main repo, the whole process matters.

kezhenxu94 (approved):
Turns out to be my local Go env problem :(

EvanLjp:
> @kezhenxu94 @EvanLjp I think this is why we use a `shell` rather than a simple make file to build src release.
> 
> @kezhenxu94 didn't do `git submodule init/update`, but run `make release-src`, then still src is not complete.

protocol files have been already added to src release.

wu-sheng:
> protocol files have been already added to src release.

Try this, use an empty folder, do git clone, then `make`. Your tar ball is broken.

wu-sheng:
You are feeling good, because you use your dev env.

PR #39, opened 2021-04-21
set MAXPROCS according to real cpu quota

wangrzneu:
GOMAXPROCS needs to be set according to real CPU quota especially in container environment with CPU limit, but runtime.NumCPU() will return the CPU number of machine, which may be very large and will affect the performance. "go.uber.org/automaxprocs" can set the GOMAXPROCS according to the environment automatically.


gxthrj (dismissed):

EvanLjp:
let me fix the ci process, please wait for a moment

gxthrj:
> let me fix the ci process, please wait for a moment

OK

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/sirupsen/logrus v1.7.0
 	github.com/spf13/viper v1.7.1
 	github.com/urfave/cli/v2 v2.3.0
+	go.uber.org/automaxprocs v1.4.0
```
wu-sheng, line 17: As this is a new runtime dependency, we need you to update the LICENSE.

wu-sheng (changes requested):
Confirming LICENSE

go.mod
wangrzneu: Done

wangrzneu:
@wu-sheng Please help to review it.

wu-sheng (approved):

PR #41, opened 2021-04-21
fixed lib version of submodule

EvanLjp:


wu-sheng (dismissed):

EvanLjp:
@wu-sheng @kezhenxu94  Because of the changing of the gRPC, the ci process often failed. Do u have any suggestions?

wu-sheng (approved):

PR #43, opened 2021-06-03
change go api package

EvanLjp:
remove  old genernate operation in CI  & change package

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ ARCH = amd64
 
 SHELL = /bin/bash
 
-all: deps verify build gen-docs check
+all: deps verify build check
 
 .PHONY: tools
 tools:
@@ -54,15 +54,6 @@ tools:
 deps: tools
 	$(GO_GET) -v -t -d ./...
 
-.PHONY: gen-codes
-gen-codes:
-	$(PROTOC) --version || sh tools/install_protoc.sh
-	/bin/sh tools/protocol_gen.sh
-
-.PHONY: gen-docs
-gen-docs: build
-	rm -rf $(PLUGIN_DOC_DIR)
-	$(OUT_DIR)/$(BINARY)-$(VERSION)-$(OSNAME)-$(ARCH) docs --output=$(PLUGIN_DOC_DIR)
 
 .PHONY: lint
 lint: tools
@@ -80,7 +71,7 @@ clean: tools
 	-rm -rf coverage.txt
 
 .PHONY: build
-build: clean gen-codes deps linux darwin windows
+build: clean deps linux darwin windows
 
 .PHONY: check
 check: clean
--- a/protocol/satellite-protocol/satellite/Event.proto
+++ b/protocol/satellite-protocol/satellite/Event.proto
@@ -20,8 +20,7 @@ syntax = "proto3";
 
 package skywalking.v3;
 
-option go_package = "satellite/protocol";
-
+option go_package = "skywalking.apache.org/repo/goapi/satellite/event/v1";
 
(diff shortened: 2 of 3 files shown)


wu-sheng (approved):

Makefile
```diff
@@ -54,15 +54,6 @@ tools:
...
 
-.PHONY: gen-codes
-gen-codes:
-	$(PROTOC) --version || sh tools/install_protoc.sh
-	/bin/sh tools/protocol_gen.sh
-
-.PHONY: gen-docs
-gen-docs: build
-	rm -rf $(PLUGIN_DOC_DIR)
-	$(OUT_DIR)/$(BINARY)-$(VERSION)-$(OSNAME)-$(ARCH) docs --output=$(PLUGIN_DOC_DIR)
```
kezhenxu94, line 65: I'm not sure why you remove these 2 targets, if you do, please also remove the related files, such as `tools/protocol_gen.sh`, as it's not used and it doesn't work now since you modify the package name.

protocol/satellite-protocol/satellite/Event.proto
```diff
@@ -20,8 +20,7 @@ syntax = "proto3";
 
 package skywalking.v3;
 
-option go_package = "satellite/protocol";
-
+option go_package = "skywalking.apache.org/repo/goapi/satellite/event/v1";
```
kezhenxu94, line 23: How does it work to just rename the package without recompile the proto files into go files?

wu-sheng:
The next step should move the dependency to `skywalking-goapi` repo, @EvanLjp gets blocked by his network issue.

Makefile
wu-sheng: One file wouldn't be able to be removed, satellite has its own event proto. But we should remove most of others, and host the generated source codes directly to avoid the complexity.

wu-sheng:
Consider adding `satellite/internal` folder into goapi repo, and move proto with generated source codes there. Is this better?

kezhenxu94:
> Consider adding `satellite/internal` folder into goapi repo, and move proto with generated source codes there. Is this better?

import paths containing `internal` is not available outside of the `goapi` module, this is restricted by Go compiler if I remember correctly

wu-sheng:
> > Consider adding `satellite/internal` folder into goapi repo, and move proto with generated source codes there. Is this better?
> 
> import paths containing `internal` is not available outside of the `goapi` module, this is restricted by Go compiler if I remember correctly

Any suggestion? My point is only hosting generated codes in `goapi` repo.

kezhenxu94:
Now the import path in this repo is `github.com/apache/skywalking-satellite/protocol/gen-codes/satellite/protocol`, I think adding `satellite` folder in `goapi` repo is enough, now

wu-sheng:
@EvanLjp Take a look.

EvanLjp:
> Now the import path in this repo is `github.com/apache/skywalking-satellite/protocol/gen-codes/satellite/protocol`, I think adding `satellite` folder in `goapi` repo is enough, now

add a proto file in go-API rather than download the proto file?

kezhenxu94:
> > Now the import path in this repo is `github.com/apache/skywalking-satellite/protocol/gen-codes/satellite/protocol`, I think adding `satellite` folder in `goapi` repo is enough, now
> 
> add a proto file in go-API rather than download the proto file?

Create a folder `satellite` in goapi repo, and add the proto files (i.e. `Event.proto`), and their generated `.go` files there. Then we import the goapi module in this repo.

EvanLjp:
okay

PR #46, opened 2021-06-04
remove-gen-codes-with-goapi

EvanLjp:
remove-gen-codes-with-goapi


kezhenxu94 (approved):

docs/en/setup/plugins/plugin-list.md
```diff
@@ -8,9 +8,9 @@
 - Fetcher
 - Filter
 - Forwarder
+	- [meter-grpc-forwarder](./forwarder_meter-grpc-forwarder.md)
 	- [nativelog-grpc-forwarder](./forwarder_nativelog-grpc-forwarder.md)
 	- [nativelog-kafka-forwarder](./forwarder_nativelog-kafka-forwarder.md)
-	- [meter-grpc-forwarder](./forwarder_meter-grpc-forwarder.md)
```
wu-sheng, line 13: Why remove this?

EvanLjp: this is auto-generated, I would check the reason

wu-sheng:
Have we removed submodule permanently?

docs/en/setup/plugins/plugin-list.md
EvanLjp: only sequence change, order by letter

kezhenxu94 (approved):

wu-sheng (approved):

PR #42, opened 2021-05-26
feat: add prometheus fetcher

gxthrj:

PR #49, opened 2021-06-04
Support generate plugin list index config in menu file

mrproliu:
* Support generate plugin list index to `menu.yml`
* Fix wrong contribute plugin docs.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -215,7 +215,6 @@ Apache 2.0 licenses
 The following components are provided under the Apache License. See project link for details.
 The text of each license is the standard Apache 2.0 license.
 
-go-yaml (yaml) v3.0.0: https://github.com/go-yaml/yaml Apache 2.0
 
 ========================================================================
 BSD licenses
--- a/LICENSE
+++ b/LICENSE
@@ -215,6 +215,7 @@ Apache 2.0 licenses
 The following components are provided under the Apache License. See project link for details.
 The text of each license is the standard Apache 2.0 license.
 
+go-yaml (yaml) v3.0.0: https://github.com/go-yaml/yaml Apache 2.0
 
 ========================================================================
 BSD licenses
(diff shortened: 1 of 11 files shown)


go.mod
```diff
@@ -15,5 +15,6 @@ require (
 	go.uber.org/automaxprocs v1.4.0
 	google.golang.org/grpc v1.38.0
 	google.golang.org/protobuf v1.26.0
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
```
kezhenxu94, line 18: Usually , a new item in `go.mod` requires a new item in `LICENSE` file

mrproliu: Thx, have added.

wu-sheng: Is this for satellite runtime? If not, we don't need the license.

LICENSE
```diff
@@ -215,6 +215,7 @@ Apache 2.0 licenses
 The following components are provided under the Apache License. See project link for details.
 The text of each license is the standard Apache 2.0 license.
 
+go-yaml (yaml) v3.0.0: https://github.com/go-yaml/yaml Apache 2.0
```
kezhenxu94, line 218: This is binary dependency, should be in `dist/LICENSE` file

go.mod
kezhenxu94: > Is this for satellite runtime? If not, we don't need the license.

It's built into the binary, although we only use it to generate docs

kezhenxu94: > > Is this for satellite runtime? If not, we don't need the license.
> 
> It's built into the binary, although we only use it to generate docs

Also, the other PR #42  will use this at runtime, I think it's ok to have this license for now, no matter which PR gets merged first.

mrproliu: Should I need to use the YAML framework v2 version? I have checked, the result basically the same.

kezhenxu94: > Should I need to use the YAML framework v2 version? I have checked, the result basically the same.

Let's use v3, I think newer version is preferable

kezhenxu94 (approved):

wu-sheng (approved):
Let's try whether this works for website tool.

PR #53, opened 2021-06-20
Support profile channel

mrproliu:
* Using none queue mode to support the SkyWalking native profile channel

### DIFF
--- a/configs/satellite_config.yaml
+++ b/configs/satellite_config.yaml
@@ -141,3 +141,27 @@ pipes:
       client_name: grpc-client
       forwarders:
         - plugin_name: nativetracing-grpc-forwarder
+  - common_config:
+      pipe_name: profilepipe
+    gatherer:
+      server_name: "grpc-server"
+      receiver:
+        plugin_name: "grpc-nativeprofile-receiver"
+      queue:
+        plugin_name: "memory-queue"
+        # The maximum buffer event size.
+        event_buffer_size: ${SATELLITE_QUEUE_EVENT_BUFFER_SIZE:5000}
+    processor:
+      filters:
+    sender:
+      fallbacker:
+        plugin_name: none-fallbacker
+      # The time interval between two flush operations. And the time unit is millisecond.
+      flush_time: ${SATELLITE_TRACINGPIPE_SENDER_FLUSH_TIME:1000}
+      # The maximum buffer elements.
+      max_buffer_size: ${SATELLITE_TRACINGPIPE_SENDER_MAX_BUFFER_SIZE:200}
+      # The minimum flush elements.
+      min_flush_events: ${SATELLITE_TRACINGPIPE_SENDER_MIN_FLUSH_EVENTS:1}
+      client_name: grpc-client
+      forwarders:
+        - plugin_name: nativeprofile-grpc-forwarder
--- a/internal/satellite/module/api/module.go
+++ b/internal/satellite/module/api/module.go
@@ -27,7 +27,8 @@ var ShutdownHookTime = time.Second * 5
 
 // Module id a custom plugin interface, which defines the processing.
 type Module interface {
-	DependencyInjection(modules ...Module)
+	// Inject the dependency modules.
+	DependencyInjection(modules ...Module) error
 	// Prepare would do some preparing workers, such build connection with external services.
 	Prepare() error
 	// Boot would start the module and return error when started failed. When a stop signal received
--- a/internal/satellite/module/api/module.go
+++ b/internal/satellite/module/api/module.go
@@ -27,8 +27,6 @@ var ShutdownHookTime = time.Second * 5
 
 // Module id a custom plugin interface, which defines the processing.
 type Module interface {
-	// Inject the dependency modules.
-	DependencyInjection(modules ...Module) error
 	// Prepare would do some preparing workers, such build connection with external services.
 	Prepare() error
 	// Boot would start the module and return error when started failed. When a stop signal received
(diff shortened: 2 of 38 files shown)


configs/satellite_config.yaml
```diff
@@ -140,4 +140,28 @@ pipes:
...
\ No newline at end of file
+        - plugin_name: nativetracing-grpc-forwarder
+  - common_config:
+      pipe_name: profilepipe
+    gatherer:
+      server_name: "grpc-server"
+      receiver:
+        plugin_name: "grpc-nativeprofile-receiver"
+      queue:
+        plugin_name: "memory-queue"
```
wu-sheng, line 151: This doesn't look like `none queue mode`. Could you be clear your meaning?

mrproliu: in the profiling protocol, we also need to use a queue to transform thread dumps.

wu-sheng: > Using none queue mode to support the SkyWalking native profile channel

What does this comment mean?

wu-sheng:
This is only a profile snapshot reporting channel, right?

mrproliu:
> This is only a profile snapshot reporting channel, right?

Also support the profile query and report finish channel, the profile channel mixes the none queue and queue data.

plugins/receiver/grpc/nativetracing/receiver.go
```diff
@@ -52,6 +53,9 @@ func (r *Receiver) RegisterHandler(server interface{}) {
 	v3.RegisterTraceSegmentReportServiceServer(r.Server, r.service)
 }
 
+func (r *Receiver) RegisterSyncProcessor(_ api.SyncProcessor) {
```
wu-sheng, line 56: This is API definition, right? It seems missing comments.

wu-sheng (approved):
@EvanLjp Please recheck.

EvanLjp:
OK，I would do review tomorrow, I just returned to Beijing today.

plugins/receiver/grpc/nativetracing/receiver.go
mrproliu: It has been add comments on the `Receiver` interface. https://github.com/apache/skywalking-satellite/pull/53/files#diff-d7d2624abe5a66c8d5e294b65af7c537db3cece721746cec346e4582614d128eR37-R38

EvanLjp:
How about adding a new module called invoker to do the sync process rather than coupling with the current gatherer, processor, and sender?

mrproliu:
> How about adding a new module called invoker to do the sync process rather than coupling with the current gatherer, processor, and sender?

Thank you for the response. Would you do more instruction?

internal/satellite/module/processor/processor.go
```diff
@@ -83,3 +84,18 @@ func (p *Processor) Boot(ctx context.Context) {
...
+
+func (p *Processor) syncProcessor(e *v1.SniffData) (*v1.SniffData, error) {
+	// build contest
+	c := &event.OutputEventContext{
+		Offset:  "",
+		Context: make(map[string]*v1.SniffData),
+	}
+	c.Put(e)
+	// processing the event with filters, that put the necessary events to OutputEventContext.
+	for _, f := range p.runningFilters {
```
EvanLjp, line 96: sync process means invoke remote method, so the lines[89-99] should be removed

internal/satellite/module/sender/sender.go
```diff
@@ -216,3 +217,12 @@ func (s *Sender) consume(batch *buffer.BatchBuffer) {
 func (s *Sender) InputDataChannel() chan<- *event.OutputEventContext {
 	return s.input
 }
+
+func (s *Sender) SyncProcess(data *v1.SniffData) (*v1.SniffData, error) {
```
EvanLjp, line 221: How about adding a new interface in forwarder API called supportedSyncInvoke to select the sync forwarder

internal/satellite/module/api/module.go
```diff
@@ -27,7 +27,7 @@ var ShutdownHookTime = time.Second * 5
 
 // Module id a custom plugin interface, which defines the processing.
 type Module interface {
-
+	DependencyInjection(modules ...Module)
```
wu-sheng, line 30: Miss comments for core level interface.

wu-sheng: If you want to run DI, how could you figure out the elements of different indices, and do the typecast?

internal/satellite/module/gatherer/fetcher_gatherer.go
```diff
@@ -117,3 +121,7 @@ func (f *FetcherGatherer) OutputDataChannel() <-chan *queue.SequenceEvent {
 func (f *FetcherGatherer) Ack(lastOffset event.Offset) {
 	f.runningQueue.Ack(lastOffset)
 }
+
+func (f *FetcherGatherer) DependencyInjection(modules ...module.Module) {
+	f.processor = modules[0].(processor.Processor)
```
wu-sheng, line 126: I can see a `[0]` directly, could you explain the DI design logic?

EvanLjp: please recheck codes.

wu-sheng:
@EvanLjp I am still confused about `DependencyInjection` method. Why do you think we need this kind of abstract way to inject dependencies? Rather than the previous ways.

wu-sheng (approved):

wu-sheng:
@EvanLjp Thanks for making clean APIs back.

PR #59, opened 2021-07-29
feat: add Prometheus service discovery in Kubernetes

gxthrj:
- [x] 0.Add Prometheus service discovery in kubernetes 
- [x] 1.fix: Add string type matching in overrideSlice
- [x] 2.fix: Initialize runningQueue in FetcherGatherer
- [x] 3.fix: Initialize OutputChannel in Prometheus fetcher plugin

### DIFF
--- a/configs/satellite_config.yaml
+++ b/configs/satellite_config.yaml
@@ -249,49 +249,49 @@ pipes:
         - plugin_name: nativemeter-grpc-forwarder
   - common_config:
       pipe_name: prometheus-fetcher
-      gatherer:
-        fetcher:
-          plugin_name: "prometheus-metrics-fetcher"
-          scrape_configs:
-            - job_name: 'prometheus'
-              metrics_path: '/metrics'
-              scrape_interval: 10s
-              static_configs:
-                - targets:
-                    - "127.0.0.1:9100"
-            - job_name: 'prometheus-k8s'
-              metrics_path: '/metrics'
-              scrape_interval: 10s
-              tls_config:
-                ca_file: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
-              bearer_token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
-              kubernetes_sd_configs:
-                - role: pod
-                  selectors:
-                    - role: pod
-                      label: "app=prometheus"
-              relabel_configs:
-                - source_labels: [ __meta_kubernetes_pod_name ]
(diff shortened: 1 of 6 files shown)


wu-sheng:
Please fix CI.

configs/satellite_config.yaml
```diff
@@ -247,3 +247,51 @@ pipes:
       client_name: grpc-client
       forwarders:
         - plugin_name: nativemeter-grpc-forwarder
+  - common_config:
```
wu-sheng, line 250: Should the fetcher pipeline be in the document only?
I think in default, we don't scratch from anywhere.

gxthrj: OK, will move to docs

configs/satellite_config.yaml
```diff
@@ -247,3 +247,51 @@ pipes:
...
+        plugin_name: "memory-queue"
+        # The maximum buffer event size.
+        event_buffer_size: ${SATELLITE_QUEUE_EVENT_BUFFER_SIZE:5000}
+    processor:
+      filters:
+    sender:
+      fallbacker:
+        plugin_name: none-fallbacker
+      # The time interval between two flush operations. And the time unit is millisecond.
+      flush_time: ${SATELLITE_LOGPIPE_SENDER_FLUSH_TIME:1000}
```
mrproliu, line 290: Please rename the environment key, is not the log pipeline.

mrproliu (commented):
Basic LGTM.

configs/satellite_config.yaml
gxthrj: instead with `SATELLITE_PROMETHEUS_*`

mrproliu (approved):

wu-sheng:
@gxthrj Do you add e2e test in this PR or in a separate one? We at least need it in this week, so we could include it in 0.2.0 release.

wu-sheng:
Changelog should be updated, https://github.com/apache/skywalking-satellite/blob/main/CHANGES.md. 

There is a format issue for bug section, please fix it too.

gxthrj:
> @gxthrj Do you add e2e test in this PR or in a separate one? We at least need it in this week, so we could include it in 0.2.0 release.

Will add e2e test in a separate one, and try my best to make it done in this week.

gxthrj:
Fix https://github.com/apache/skywalking/issues/7436

PR #60, opened 2021-08-17
docs: add prometheus fetcher documents

gxthrj:
- [x] Remove the Prometheus fetcher configuration from the default configuration
- [x] Modify document of Prometheus fetcher


docs/en/setup/plugins/fetcher_prometheus-metrics-fetcher.md
```diff
@@ -1,13 +1,34 @@
 # Fetcher/prometheus-metrics-fetcher
+
```
wu-sheng, line 2: @mrproliu This should be generated by the comments?

gxthrj: Yes, I used `make gen-docs` instead.

wu-sheng (approved):

mrproliu (approved):

PR #64, opened 2021-09-13
Polishing documentations

mrproliu:
1. Add default settings description
2. Add some examples to help users quick start
3. Add support forwarder config on fetcher and receiver for document
4. Remove some links on the menu.yaml
5. Fix some wrong package name

### DIFF
--- a/docs/en/setup/README.md
+++ b/docs/en/setup/README.md
@@ -3,8 +3,6 @@ First and most important thing is, SkyWalking Satellite startup behaviours are d
 
 ## Requirements and default settings
 
-Requirement: **>= Go 1.14 are tested**. Other versions are not tested and may or may not work.
-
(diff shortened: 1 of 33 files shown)


docs/en/setup/README.md
```diff
@@ -1,11 +1,29 @@
 # Setup
 First and most important thing is, SkyWalking Satellite startup behaviours are driven by configs/satellite_config.yaml. Understanding the setting file will help you to read this document.
 
+## Requirements and default settings
+
+Requirement: **>= Go 1.14 are tested**. Other versions are not tested and may or may not work.
```
kezhenxu94, line 6: Go binaries don't relies on Go version, this requirement is for development only I think.

mrproliu: You're right, has been deleted.

wu-sheng (approved):

PR #68, opened 2021-09-22
Support load-balance GRPC client with static server list

mrproliu:
Follow https://github.com/apache/skywalking/issues/7247
Using customize GRPC resolver to implement, and using robin load-balancer to select connect on each request

### DIFF
--- a/plugins/client/grpc/client.go
+++ b/plugins/client/grpc/client.go
@@ -28,14 +28,16 @@ import (
 	"github.com/apache/skywalking-satellite/internal/pkg/config"
 	"github.com/apache/skywalking-satellite/internal/pkg/log"
 	"github.com/apache/skywalking-satellite/plugins/client/api"
+	"github.com/apache/skywalking-satellite/plugins/client/grpc/resolvers"
 )
 
 const Name = "grpc-client"
 
 type Client struct {
 	config.CommonFields
+	// server finder config
+	ServerFinderConfig resolvers.ServerFinderConfig `mapstructure:",squash"`
 	// config
-	ServerAddr         string `mapstructure:"server_addr"`          // The gRPC server address
 	EnableTLS          bool   `mapstructure:"enable_TLS"`           // Enable TLS connect to server
 	ClientPemPath      string `mapstructure:"client_pem_path"`      // The file path of client.pem. The config only works when opening the TLS switch.
 	ClientKeyPath      string `mapstructure:"client_key_path"`      // The file path of client.key. The config only works when opening the TLS switch.
@@ -62,7 +64,7 @@ func (c *Client) Description() string {
 
 func (c *Client) DefaultConfig() string {
 	return `
-# The gRPC server address (default localhost:11800). 
+# The gRPC server address (default localhost:11800), multiple addresses are split by ",".
 server_addr: localhost:11800
 
 # The TLS switch (default false).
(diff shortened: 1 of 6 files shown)


wu-sheng:
Could you be clear about what is the balance policy?

wu-sheng:
Or this is just for loading multiple-servers list?

plugins/client/grpc/client.go
```diff
@@ -100,8 +102,15 @@ func (c *Client) Prepare() error {
...
+	// server address resolver
+	resolvers.RegisterAllGrpcResolvers()
+
 	// connect to server
-	client, err := grpc.Dial(c.ServerAddr, *cfg...)
+	target, err := resolvers.BuildTarget(&c.ServerFinderConfig)
+	if err != nil {
+		return fmt.Errorf("cannot build grpc target: %v", err)
+	}
+	client, err := grpc.Dial(target, *cfg...)
```
wu-sheng, line 113: This makes me feel you have one logic client reference? Who controls the routine rule?

plugins/client/grpc/client_config.go
```diff
@@ -73,6 +73,9 @@ func (c *Client) loadConfig() (*[]grpc.DialOption, error) {
 		return err
 	})
 
+	// using round-robin load balancer
+	options = append(options, grpc.WithDefaultServiceConfig(`{"loadBalancingPolicy":"round_robin"}`))
```
wu-sheng, line 77: Is this the routine rule for https://github.com/apache/skywalking-satellite/pull/68/files#r714050900?

mrproliu:
> Or this is just for loading multiple-servers list?

Yes, just load-balance all existing multiple servers when sending requests.

plugins/client/grpc/client_config.go
wu-sheng: If so, it is OK for now, but we will need our own LB-policy implementation later. Because this is not real protocol-oriented `round_robin`.

plugins/client/grpc/client.go
```diff
@@ -100,8 +102,15 @@ func (c *Client) Prepare() error {
 		"client_name": Name,
 	})})
 
+	// server address resolver
+	resolvers.RegisterAllGrpcResolvers()
+
 	// connect to server
-	client, err := grpc.Dial(c.ServerAddr, *cfg...)
+	target, err := resolvers.BuildTarget(&c.ServerFinderConfig)
```
wu-sheng, line 109: Notice, this target could be changed dynamically. Such as in k8s, a domain name of services could have dynamic list of pods IP eventually. We need to resolve this case.

plugins/client/grpc/client_config.go
mrproliu: That's true. I think it is enough for us to specify a load-balance method.

plugins/client/grpc/client.go
wu-sheng: This is one of the key issue, ALS client can't dynamically routine requests to all OAP backends.
We need to get the list periodically, and make sure the list getting updated, meanwhile all data(traces/logs/etc.) gets to be sent through one channel.

wu-sheng (approved):
For static server list, this is fine.

plugins/client/grpc/client.go
mrproliu: Yes, For the k8s case, there just save the Kubernetes pod/service config(by label selector), not include real address config. The static address config is a special case because we just need to specify the address list.
We use resolver to update the server address list. Please take a look on: https://github.com/apache/skywalking-satellite/pull/68/files#diff-31cdd706c94e7a14dab4bf57861b31fc5b7e919d946fd480c593f3f18f8aafd4R69-R76

PR #69, opened 2021-09-24
Support load-balance GRPC client with Kubenetes selector

mrproliu:
Follow https://github.com/apache/skywalking/issues/7247
I found the Kubernetes selector is similar to the [Prometheus Kubernetes config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config). So I wrapped and use analyze the data to get the addresses.

I found the plugin could not auto-generate the plugin settings, only has default settings, but some options config are not declared in the documentation. 
So I will create other PR to generate them by config field comment.

### DIFF
--- a/docs/en/setup/examples/grpc-load-balance-client/README.md
+++ b/docs/en/setup/examples/grpc-load-balance-client/README.md
@@ -1,6 +1,6 @@
 # GRPC load balance client
 
-GRPC client support connect to multiple server address, and using round-robin to load-balance server before send each request.
+GRPC client support connect to multiple server address, and use `round-robin` policy for load-balance server before send each request.
 
 ## Server Discovery
 
@@ -28,7 +28,7 @@ sharing:
       insecure_skip_verify: ${SATELLITE_GRPC_INSECURE_SKIP_VERIFY:false}
       # The file path oca.pem. The config only works when opening the TLS switch.
       ca_pem_path: ${SATELLITE_grpc_CA_PEM_PATH:"ca.pem"}
-      # How frequently to check the connection
+      # How frequently to check the connection(second)
       check_period: ${SATELLITE_GRPC_CHECK_PERIOD:5}
       # The auth value when send request
       authentication: ${SATELLITE_GRPC_AUTHENTICATION:""}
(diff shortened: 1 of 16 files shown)


docs/en/setup/examples/grpc-load-balance-client/README.md
```diff
@@ -0,0 +1,84 @@
+# GRPC load balance client
+
+GRPC client support connect to multiple server address, and using round-robin to load-balance server before send each request.
```
wu-sheng, line 3:
```suggestion
GRPC client support connect to multiple server address, and use `round-robin` policy for load-balance.
```
Does this base on per streaming or per message?

docs/en/setup/examples/grpc-load-balance-client/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Server Discovery
+
+Support two ways to locate the server list:
+1. Static server list: Define the server address list.
```
wu-sheng, line 8: Why do we need to mention `static list again`? I remember we used to have one already.

docs/en/setup/examples/grpc-load-balance-client/README.md
```diff
@@ -0,0 +1,84 @@
...
+      enable_TLS: ${SATELLITE_GRPC_ENABLE_TLS:false}
+      # The file path of client.pem. The config only works when opening the TLS switch.
+      client_pem_path: ${SATELLITE_GRPC_CLIENT_PEM_PATH:"client.pem"}
+      # The file path of client.key. The config only works when opening the TLS switch.
+      client_key_path: ${SATELLITE_GRPC_CLIENT_KEY_PATH:"client.key"}
+      # InsecureSkipVerify controls whether a client verifies the server's certificate chain and host name.
+      insecure_skip_verify: ${SATELLITE_GRPC_INSECURE_SKIP_VERIFY:false}
+      # The file path oca.pem. The config only works when opening the TLS switch.
+      ca_pem_path: ${SATELLITE_grpc_CA_PEM_PATH:"ca.pem"}
+      # How frequently to check the connection
```
wu-sheng, line 75: Add unit please.

docs/en/setup/examples/grpc-load-balance-client/README.md
mrproliu: Here is an example to let users know how to define multiple servers, `static server lists` and `Kubernetes selector` are two different ways.

docs/en/setup/examples/grpc-load-balance-client/README.md
mrproliu: Per streaming.

docs/en/setup/examples/grpc-load-balance-client/README.md
wu-sheng: I see. Then you missed the menu update.

mrproliu: Added.

wu-sheng:
> I found the plugin could not auto-generate the plugin settings, only has default settings, but some options config are not declared in the documentation.

Noticed this too. Good catch.

wu-sheng (approved):

PR #74, opened 2021-10-09
Support Envoy ALS v2/v3

mrproliu:
Follow https://github.com/apache/skywalking/issues/7774, Support transmit Envoy ALS V2 and V3 protocol.

I found the satellite has the `protobuf` register conflict problem, the `Prometheus fethcer` and `goapi` both need Prometheus metrics protocol, so I use `google.golang.org/protobuf/reflect/protoregistry.conflictPolicy=ignore` to ignore the conflict message. If anyone has another good idea, please let me know.

### DIFF
--- a/configs/satellite_config.yaml
+++ b/configs/satellite_config.yaml
@@ -254,9 +254,9 @@ pipes:
       receiver:
         plugin_name: "grpc-envoy-als-v2-receiver"
         # The time interval between two flush operations. And the time unit is millisecond.
-        flush_time: ${SATELLITE_ENVOYALSV2PIPE_RECEIVER_FLUSH_TIME:1000}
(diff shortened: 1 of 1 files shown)


wu-sheng:
If conflict exists, which dependency is being used?

mrproliu:
> If conflict exists, which dependency is being used?

Currently is using `goapi`.

mrproliu:
> Please copy this test to the main repository, to replace the ALS E2E.
> 
> @kezhenxu94 Please recheck.

Ok. Current this E2E is only set up and verify the protocol transmit function, I need to set up a complete environment and add more verify if add to the main repo. 

Also, Should I need to include satellite service into the main repo ALS E2E?

PR #75, opened 2021-10-12
fix: errors when converting meter data from histogram and summary

gxthrj:
This PR fix https://github.com/apache/skywalking/issues/7892
- [x] The labels in the meter are not right.
- [x] The *_sum metrics are missing.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -27,7 +27,7 @@ import (
 )
 
 // version will be initialized when building
-var version string = "lastest"
+var version = "latest"
 
 func main() {
 	app := cli.NewApp()
--- a/docs/en/concepts-and-designs/module_design.md
+++ b/docs/en/concepts-and-designs/module_design.md
@@ -43,7 +43,7 @@ LifeCycle
 
 ## Plugins
 
-Plugin is the minimal components in the module. Sateliite has 2 plugin catalogs, which are sharing plugins and normal plugins.
+Plugin is the minimal components in the module. Satellite has 2 plugin catalogs, which are sharing plugins and normal plugins.
 
 - a sharing plugin instance could be sharing with multiple modules in the different pipes.
 - a normal plugin instance is only be used in a fixed module of the fixed pipes.
@@ -102,7 +102,7 @@ There are 7 kinds of normal plugins in Satellite, which are Receiver, Fetcher, Q
                                             | | A |     | I |    |Segment Forwarder|   |
                                             | | T |     | S |    |    (Fallbacker) |   |
                                             | | C |     | P |     -----------------    |
-                                            | | H |  => | A |                          | ===> Kakfa/OAP
+                                            | | H |  => | A |                          | ===> Kafka/OAP
                                             | | B |     | T | =>        ......         |
                                             | | U |     | C |                          |
                                             | | F |     | H |     -----------------    |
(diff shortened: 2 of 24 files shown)


gxthrj:
> Is there an issue to link or close?

Sorry, there is no issue related before. Should I create an issue for this?

wu-sheng:
You don't have to. But please update the change log.

wu-sheng:
We really need Prometheus e2e for covering this case. Currently, it seems we are testing manually every time.

gxthrj:
> Hi @gxthrj , I tested my case and found a mirror issue. There is a `+Inf` bucket in Histogram, then satellite would convert it to a numeric. Like `-9.223372036854776E18`. That causes OAP exception.

That is the difference between `skywalking` and `prometheus`.

For example
In Prometheus , get histogram
```
# HELP http_request_duration_seconds A histogram of the request duration.
# TYPE http_request_duration_seconds histogram
http_request_duration_seconds_bucket{le="0.05"} 1100
http_request_duration_seconds_bucket{le="0.5"} 1600
http_request_duration_seconds_bucket{le="1"} 2100
http_request_duration_seconds_bucket{le="+Inf"} 2600
http_request_duration_seconds_sum 5050
http_request_duration_seconds_count 2600
```
When translating to Meter in `skywalking`, it will be like this 
```
http_request_duration_seconds_count [] 2500
http_request_duration_seconds_sum [] 5000
http_request_duration_seconds [] [bucket:-9.223372036854776e+18  count:1000 bucket:0.05  count:1500 bucket:0.5  count:2000 bucket:1  count:2500]
```

The representation of the bucket is different.

In proemtheus, the first bucket is `le=0.05` histogram if from <= 0.05 in this example. -- (minNum, 0.05 ]
In skywalking, the first bucket is the minInt64, the second bucket is 0.05.  -- (minNum, 0.05]
But they are the same.

wu-sheng:
> There is a +Inf bucket in Histogram, then satellite would convert it to a numeric. Like -9.223372036854776E18.

The minimal number is translating to this, but I am not sure whether it is necessary(@gxthrj ), or what is the exception you facing? (@dmsolr )

gxthrj:
I think it is necessary.
I can use math.Inf(-1) instead of minInt

wu-sheng:
> use math.Inf(-1) instead of minInt

Does anyone verify this in the backend? I don't think this could be changed randomly.

gxthrj:
> > use math.Inf(-1) instead of minInt
> 
> Does anyone verify this in the backend? I don't think this could be changed randomly.

OAP can read this. @dmsolr

dmsolr:
> > > ```
> > >     /**
> > >      * The element in the buckets represent the minimal value of this bucket, the max is defined by the next element.
> > >      * Such as 0, 10, 50, 100 means buckets are [0, 10), [10, 50), [50, 100), [100, infinite+).
> > >      *
> > >      * The {@link Long#MIN_VALUE} could be the first bucket element to indicate there is no minimal value.
> > >      */
> > >     private long[] buckets;
> > > ```
> > > This is bucket definition. @gxthrj I think what you posted is not mapping to it.
> > 
> > 
> > I refer to this description: https://github.com/apache/skywalking-goapi/blob/main/collect/language/agent/v3/Meter.pb.go#L103-L107
> 
> No matter what @dmsolr is facing about bug in the OAP, it should be fixed.
> 
> My point is, -Inf is really necessary? Because histogram in SkyWalking is not expected to have such small value. It is better to set minimal value as 0 or user setting.

Yeap. In fact, `-Inf/+Inf` is a symbol, so I wanna confirm whether Satellite keeps this symbol. If not, we need to define a key.

wu-sheng:
I talked with @mrproliu. The `infinite-` as bucket is not supported actually. We will update the proto and implement this in the OAP first.

PR #77, opened 2021-10-17
Use 8.9.0-dev for Satellite testing.

wu-sheng:


wu-sheng:
@mrproliu It seems the latest image makes the test failing. Please recheck. Others are correct, only ALS cases fail.

PR #80, opened 2021-10-31
Enhance Envoy protocols E2E

mrproliu:
Using `KinD` environment to set up the `Istio` for testing envoy ALS and Metrics protocol


.github/workflows/e2e.yaml
```diff
@@ -34,26 +34,6 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: true
-      - uses: apache/skywalking-infra-e2e@d37840fb22f7c78f03ab5932bbdccf925cb5ef73
+      - uses: apache/skywalking-infra-e2e@main
```
wu-sheng, line 37: Please update commit id rather than `main.

.github/workflows/e2e-istio.yaml
```diff
@@ -0,0 +1,86 @@
...
+        with:
+          go-version: 1.16
+        id: go
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Build Docker Image
+        shell: bash
+        run: make docker
+      - uses: apache/skywalking-infra-e2e@main
```
wu-sheng, line 57: Same here.

wu-sheng (approved):

PR #82, opened 2021-11-09
Support partition queue

mrproliu:
* Support partition queue. Resolve https://github.com/apache/skywalking/issues/8077
* Add self-telemetry to the `ALS` protocol and `queue`.


configs/satellite_config.yaml
```diff
@@ -103,6 +105,8 @@ pipes:
         plugin_name: "memory-queue"
         # The maximum buffer event size.
         event_buffer_size: ${SATELLITE_QUEUE_EVENT_BUFFER_SIZE:5000}
+        # The partition count of queue.
+        partition_count: ${SATELLITE_QUEUE_PARTITION_COUNT:1}
```
wu-sheng, line 109: Remove count. Partition is enough.

docs/en/setup/examples/high-throughput-queue/README.md
```diff
@@ -0,0 +1,19 @@
+# High throughput queue
+
+High throughput queues can send messages to multiple channels using by `round-robin` policy for load-balance, and consume messages from each queue in parallel.
+This prevents the single-threaded sender from performing network IO, blocking the `recieve`/`fetch`/`consumer` to the queue.
```
wu-sheng, line 4: Receiver and fetcher.

docs/en/setup/plugins/queue_mmap-queue.md
```diff
@@ -15,6 +15,9 @@ flush_period: 1000
 flush_ceiling_num: 10000
 # The max size of the input event. Default value is 20k.
 max_event_size: 20480
+
+# The partition count of queue.
+partition_count: 1
```
wu-sheng, line 20: mmap queue supports partitions?

mrproliu: Yes, It supports. I wrap a partitioned queue to manage all sub-queue(memory or mmap).

docs/en/setup/examples/high-throughput-queue/README.md
```diff
@@ -0,0 +1,19 @@
+# High throughput queue
+
+High throughput queues can send messages to multiple channels using by `round-robin` policy for load-balance, and consume messages from each queue in parallel.
```
kezhenxu94, line 3:
```suggestion
High throughput queues can send messages to multiple channels by using `round-robin` policy for load-balance, and consume messages from each queue in parallel.
```

docs/en/setup/examples/high-throughput-queue/README.md
```diff
@@ -0,0 +1,19 @@
...
+This prevents the single-threaded sender from performing network IO, blocking the `recieve`/`fetch`/`consumer` to the queue.
+
+## Config
+
+Both existing two queues(`memory`, `mmap`) support this future, add the `partition_count` to the queue config, default is `1`.
```
kezhenxu94, line 8:
```suggestion
Both existing two queues(`memory`, `mmap`) support this feature, add the `partition_count` to the queue config, default is `1`.
```

plugins/queue/mmap/queue_operation.go
```diff
@@ -20,11 +20,14 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 
+//go:build !windows
```
kezhenxu94, line 23: What's the difference?

```
//go:build !windows
// +build !windows
```

mrproliu: They are the same, I'm using the go@1.17 version in my local machine. So It auto-generates by `go fmt`, being deleted.

docs/menu.yml
```diff
@@ -27,13 +27,13 @@ catalog:
               path: /en/concepts-and-designs/project-goals
         - name: Design
           catalog:
-            - name: Module Design
+            - name: Module design
```
wu-sheng, line 30: I think we should keep upper cases for every noun.

wu-sheng (approved):

wu-sheng (approved):

PR #86, opened 2021-11-22
Setup full configuration, each config could be setup from environment

mrproliu:
Mostly is update the gRPC client configuration

### DIFF
--- a/configs/satellite_config.yaml
+++ b/configs/satellite_config.yaml
@@ -37,8 +37,55 @@ telemetry:
 sharing:
   clients:
     - plugin_name: "grpc-client"
+      # The gRPC server address finder type
+      finder_type: ${SATELLITE_GRPC_CLIENT_FINDER:static}
       # The gRPC server address (default localhost:11800).
       server_addr: ${SATELLITE_GRPC_CLIENT:127.0.0.1:11800}
+      # The gRPC kubernetes server address finder
+      kubernetes_config:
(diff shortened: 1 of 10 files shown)


configs/satellite_config.yaml
```diff
@@ -37,8 +37,55 @@ telemetry:
 sharing:
   clients:
     - plugin_name: "grpc-client"
+      # The gRPC server address finder type
+      finder_type: ${SATELLITE_GRPC_CLIENT_FINDER:static}
```
wu-sheng, line 41: It is better to list all options here.

docs/en/setup/plugins/client_grpc-client.md
```diff
@@ -30,6 +45,7 @@ check_period: 5
 ## Configuration
 |Name|Type|Description|
 |----|----|-----------|
+| finder_type | string | The gRPC server address finder type |
```
wu-sheng, line 48: Please describe how to activate different finders.

wu-sheng:
I remember you used to add k8s service discovery, do I remember wrongly or something?

wu-sheng:
OK, then, please make the config more clear about how to activate k8s upstream service discovery.

docs/en/setup/plugins/client_grpc-client.md
mrproliu: been added to the example file.

wu-sheng (approved):

PR #94, opened 2021-12-14
Compatible with proto.MessageV1

2zturn:


wu-sheng:
Could you be more clear about what is this about? What V1 message?

plugins/server/grpc/server_codec.go
```diff
@@ -41,28 +42,42 @@ func NewOriginalData(data []byte) *OriginalData {
...
 		return nil
 	}
-	return proto.Unmarshal(data, vv)
+
+	vv, ok := v.(proto.Message)
+	if ok {
+		return proto.Unmarshal(data, vv)
+	}
+
+	err := proto.Unmarshal(data, pbv1.MessageV2(v))
```
mrproliu, line 75: I think It needs to use `MessageV1` method?

plugins/server/grpc/server_codec.go
```diff
@@ -41,28 +42,42 @@ func NewOriginalData(data []byte) *OriginalData {
...
 		return original.Content, nil
 	}
-	return proto.Marshal(vv)
+
+	vv, ok := v.(proto.Message)
+	if ok {
+		return proto.Marshal(vv)
+	}
+
+	b, err := pbv1.Marshal(pbv1.MessageV1(v))
```
mrproliu, line 55: Please use the cast to check again.

mrproliu (commented):
Thanks for your contribution. Please fix the comment and add the changes to the [CHANGES.md](https://github.com/apache/skywalking-satellite/blob/main/CHANGES.md).

plugins/server/grpc/server_codec.go
2zturn: in my case, I use MessageV1 at first, but it doesn't work,  when change it  to MessageV2, it's work, so ...

mrproliu:
@2zturn Could you help when we need to use the V1 message? When will `V1` be needed?

2zturn:
when it works with: go.opentelemetry.io/collector/model/otlpgrpc,  this package will return the MessageV1, then stream it, the origin marshal is failed

wu-sheng:
> when it works with: go.opentelemetry.io/collector/model/otlpgrpc, this package will return the MessageV1, then stream it, the origin marshal is failed

I am a little confusing. What gRPC service do you use to forward data to Satellite?

2zturn:
> > when it works with: go.opentelemetry.io/collector/model/otlpgrpc, this package will return the MessageV1, then stream it, the origin marshal is failed
> 
> I am a little confusing. What gRPC service do you use to forward data to Satellite?

I pull data from Satellite not forward to it

wu-sheng:
`Pull from satellite` to `where`?

JaredTan95:
> `Pull from satellite` to `where`? Which format is compatible with OTEL collector? 

convert skywalking segment to otlp tracing

wu-sheng:
> > `Pull from satellite` to `where`? Which format is compatible with OTEL collector?
> 
> convert skywalking segment to otlp tracing

But this code is adding to our gRPC server, shouldn't this be added into OTEL collector? I think I need more information up to your proposal

2zturn:
ok, i will try to add into OTEL collector, thx for your all

PR #95, opened 2021-12-16
Support satellite metrics service, add the resource limiter for gRPC connect

mrproliu:
* Remove `prometheus-server` and use telemetry metrics exporter to config.
* Add the resource limiter when gRPC server accepts the connection, protect satellite resource.
* Fix the native meter protocol message load balance bug.
* Update the documentation on gRPC server config.
* Update the Licenses.
* Update E2E version and components.


docs/en/setup/plugins/forwarder_nativemeter-grpc-forwarder.md
```diff
@@ -6,4 +6,6 @@ This is a synchronization meter grpc forwarder with the SkyWalking meter protoco
 ## Configuration
 |Name|Type|Description|
 |----|----|-----------|
+| upstream_cache_size | int | The upstream cache max size |
+| upstream_cache_second | int | The upstream cache time(second) on each service instance |
```
wu-sheng, line 10: What is `cache time`? I think you mean `upstream_cache_ttl_last_used`?

And what happens if `upstream_cache_size` is reached, but no data could be removed by `upstream_cache_ttl_last_used` policy?

docs/menu.yml
```diff
@@ -153,8 +153,6 @@ catalog:
                   path: /en/setup/plugins/server_grpc-server
                 - name: HTTP Server
                   path: /en/setup/plugins/server_http-server
-                - name: Prometheus Server
-                  path: /en/setup/plugins/server_prometheus-server
```
wu-sheng, line 157: You removed this one, where is the new one?

internal/satellite/telemetry/metricsservice/base.go
```diff
@@ -0,0 +1,262 @@
+// Licensed to Apache Software Foundation (ASF) under one or more contributor
```
wu-sheng, line 1: I think `metricsservice` should be `metricservice?

docs/menu.yml
wu-sheng: I think we need a `telemetry export` document.

docs/en/setup/plugins/server_grpc-server.md
```diff
@@ -15,6 +15,12 @@ max_concurrent_streams: 32
 tls_cert_file: ""
 # The TLS key file path.
 tls_key_file: ""
+# To Accept Connection Limiter when reach the resource
+accept_limit:
+  # The max CPU utilization limit
+  cpu_utilization: 90
```
wu-sheng, line 21: I think `75%` is a more common default setting for CPU load. And have you verified this would be over 100% in multiple cores?

docs/en/setup/plugins/client_grpc-client.md
```diff
@@ -46,8 +46,8 @@ check_period: 5
 |Name|Type|Description|
 |----|----|-----------|
 | finder_type | string | The gRPC server address finder type, support "static" and "kubernetes" |
-| server_addr | string | The gRPC server address |
-| kubernetes_config | *resolvers.KubernetesConfig | The kubernetes config to lookup addresses |
+| server_addr | string | The gRPC server address, only works on "static" address finder |
+| kubernetes_config | *resolvers.KubernetesConfig | The kubernetes config to lookup addresses, only works on "kubernetes" address finder |
```
wu-sheng, line 50:
```suggestion
| server_addr | string | The gRPC server address, only works for "static" address finder |
| kubernetes_config | *resolvers.KubernetesConfig | The kubernetes config to lookup addresses, only works for "kubernetes" address finder |
```

configs/satellite_config.yaml
```diff
@@ -32,6 +32,20 @@ telemetry:
...
   instance: ${SATELLITE_TELEMETRY_SERVICE:satellite-instance}
+  # Telemetry export type, support "prometheus", "metrics_service" or "none"
+  export_type: ${SATELLITE_TELEMETRY_EXPORT_TYPE:metrics_service}
+  # Export to prometheus configuration, only works on "prometheus" exporter
+  prometheus:
+    # The prometheus server address.
+    address: ${SATELLITE_TELEMETRY_PROMETHEUS_ADDRESS::1234}
+    # The prometheus server metrics endpoint.
+    endpoint: ${SATELLITE_TELEMETRY_PROMETHEUS_ENDPOINT:/metrics}
+  # Export to MetricsService configuration, only works on "metrics_service" exporter
```
wu-sheng, line 43:
```suggestion
  # Export telemetry data through native meter format to OAP backend, only works on "metrics_service" exporter
```

configs/satellite_config.yaml
```diff
@@ -32,6 +32,20 @@ telemetry:
   service: ${SATELLITE_TELEMETRY_SERVICE:satellite-service}
   # The minimum running unit, such as the pod concept in the Kubernetes.
   instance: ${SATELLITE_TELEMETRY_SERVICE:satellite-instance}
+  # Telemetry export type, support "prometheus", "metrics_service" or "none"
+  export_type: ${SATELLITE_TELEMETRY_EXPORT_TYPE:metrics_service}
+  # Export to prometheus configuration, only works on "prometheus" exporter
```
wu-sheng, line 37:
```suggestion
  # Export telemetry data through Prometheus server, only works on "export_type=prometheus".
```

docs/en/setup/plugins/forwarder_nativemeter-grpc-forwarder.md
mrproliu: > What is cache time? I think you mean upstream_cache_ttl_last_used?

Yes, this means the maximum survival time of data when the instance is no longer sending upstream.

> And what happens if upstream_cache_size is reached, but no data could be removed by upstream_cache_ttl_last_used policy?

It's a kind of LRU Map, if the cache size is reached, it will remove the longest unused data.

docs/en/setup/plugins/server_grpc-server.md
mrproliu: It works too. I have combined all the CPU information into one utilization.

docs/en/setup/plugins/forwarder_nativemeter-grpc-forwarder.md
wu-sheng: OK, then let's rename both to `upstream_lru_cache_size` and `pstream_lru_cache_ttl`

docs/menu.yml
mrproliu: Add to a new example.

docs/en/setup/examples/feature/telemetry-exporter/README.md
```diff
@@ -0,0 +1,56 @@
+# Telemetry Exporter
+
+Satellite Telemetry Support two ways to export: `prometheus`, `metrics-service`.
```
wu-sheng, line 3:
```suggestion
Satellite supports two ways to export its own telemetry data, `prometheus` or `metrics-service`.
```

docs/en/setup/plugins/forwarder_nativemeter-grpc-forwarder.md
```diff
@@ -2,8 +2,15 @@
...
+# The upstream LRU Cache max size
+upstream_lru_cache_size: 5000
+# The upstream LRU Cache time(second) on each service instance
+upstream_lru_cache_ttl: 180
+```
 ## Configuration
 |Name|Type|Description|
 |----|----|-----------|
+| upstream_lru_cache_size | int | The upstream LRU Cache max size |
+| upstream_lru_cache_ttl | int | The upstream LRU Cache time(second) on each service instance |
```
wu-sheng, line 15: I am reading again, feel like these two names are misleading. I think we are not just caching `upstream`, the real data in the cache is the `routine_rule` of meter data.

```suggestion
| routine_rule_lru_cache_size | int | The LRU policy cache size for hosting routine rules of service instance. |
| routine_rule_lru_cache_ttl | int | The TTL of the LRU cache size for hosting routine rules of service instance. |
```

wu-sheng (approved):
Once e2e passed, I am good.

PR #98, opened 2021-12-23
Fix CVE-2021-41103, CVE-2021-3121

mrproliu:
* Update github.com/containerd/containerd to v1.5.8 to fix [CVE-2021-41103](https://avd.aquasec.com/nvd/cve-2021-41103/)
* Update github.com/gogo/protobuf to v1.3.2 to fix [CVE-2021-3121](https://avd.aquasec.com/nvd/cve-2021-3121/)
* Need to update the Go version to 1.17 when update the github.com/containerd/containerd


.github/workflows/e2e-istio.yaml
```diff
@@ -18,19 +18,16 @@ name: E2E
...
 on:
   pull_request:
-    paths:
-      - '**'
-      - '!**.md'
-  schedule:
-    - cron: '0 18 * * *'
+  push:
+    branches:
+      - main
```
wu-sheng, line 23: I think is unexpected change?

.github/workflows/build-and-test.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
-        go-version: [ 1.14, 1.15 ]
+        go-version: [ 1.17 ]
```
wu-sheng, line 31: You are upgrading the JDK version, are 1.14 and 1.15 still supported?

mrproliu: Not support anymore.

wu-sheng: Then, do we have any compiling doc or something to update?

mrproliu: Add the Go version limit in the How-to-compile.md.

wu-sheng (approved):

PR #108, opened 2022-04-08
Fix Metadata lost in the Native Meter protocol

mrproliu:
Package the data when receiving streaming meter data, and use `collectBatch` to send them.
Ignore the old transform data(if using the `mmap` queue), and log it to ignore them.

### DIFF
--- a/plugins/forwarder/grpc/nativemeter/forwarder.go
+++ b/plugins/forwarder/grpc/nativemeter/forwarder.go
@@ -100,11 +100,12 @@ func (f *Forwarder) Forward(batch event.BatchEvents) error {
 	}()
 	oldTransformDataCount := 0
 	for _, e := range batch {
-		if data, ok := e.GetData().(*v1.SniffData_MeterCollection); ok {
+		switch data := e.GetData().(type) {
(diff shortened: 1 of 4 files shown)


plugins/forwarder/grpc/nativemeter/forwarder.go
```diff
@@ -98,22 +98,28 @@ func (f *Forwarder) Forward(batch event.BatchEvents) error {
...
-			}
-		}
-		if data, ok := e.GetData().(*v1.SniffData_MeterCollection); ok {
+		switch data := e.GetData().(type) {
+		case *v1.SniffData_MeterCollection:
 			if err := f.handleMeterCollection(data, streamMap); err != nil {
 				return err
 			}
+		case *v1.SniffData_Meter:
+			oldTransformDataCount++
```
wu-sheng, line 109: As we have moved to `MeterCollection`, let's ignore meter(s) in queue(mmap) directly.

mrproliu: updated.

wu-sheng (approved):

PR #110, opened 2022-05-09
Support transmit the eBPF Process and Profiling protocol

mrproliu:
* Support transmit the eBPF Process and Profiling protocol
* Add the E2E related to the eBPF protocol
* Upgrade the oldest E2E testing
* Change the name of plugin that is not well-named


docs/en/setup/plugins/forwarder_nativeebpfprofiling-grpc-forwarder.md
```diff
@@ -0,0 +1,9 @@
+# Forwarder/nativeebpfprofiling-grpc-forwarder
```
wu-sheng, line 1: How about changing to `native/ebpfprofiling`? The word seems too long to read.

mrproliu: Rename to `native-ebpfprofiling-grpc-forward`

wu-sheng (approved):

PR #113, opened 2022-05-17
Remove the `LICENSE.tpl` in binary tar file

mrproliu:
Forget to delete the `LICENSE.tpl` in the release script


wu-sheng:
This is our own file, so, it is not an issue if we keep it in the release.

wu-sheng (approved):

mrproliu:
Yes, but it's just a template file, it's not necessary to keep it in the release binary file. So we could delete it.

PR #114, opened 2022-05-28
Downgrade unnecessary log level

mrproliu:
We have a lot of logs that don't need to be printed, change them to debug level.


mrproliu:
Change the log level in the configs looks more correct, so close this PR.

PR #116, opened 2022-07-28
Fix the missing return data when receive metrics in batch mode.

mrproliu:


CHANGES.md
```diff
@@ -7,8 +7,8 @@ Release Notes.
 #### Features
 
 #### Bug Fixes
-* Fix metadata messed up when transferring Log data.
+* Fix the missing return data when receive metrics in batch mode.
 
 #### Issues and PR
-- All issues are [here](https://github.com/apache/skywalking/milestone/137?closed=1)
+- All issues are [here](https://github.com/apache/skywalking/milestone/143?closed=1)
 - All and pull requests are [here](https://github.com/apache/skywalking-satellite/pulls?q=is%3Apr+milestone%3A1.1.0+is%3Aclosed)
```
wu-sheng, line 14: I think you need to update this to adopt version 1.0.2

wu-sheng (approved):

PR #118, opened 2022-10-08
Fix CVE-2022-21698, CVE-2022-27664, Upgrade to Go1.18

mrproliu:


wu-sheng:
Welcome back.

wu-sheng (approved):

docs/en/guides/compile/How-to-compile.md
```diff
@@ -2,7 +2,7 @@
 
 ## Go version
 
-Go version `1.17` or higher is supported for compilation.
+Go version `1.18` or higher is supported for compilation.
```
kezhenxu94, line 5: Can you confirm this is correct? Usually I found higher Go version cannot compile lower Go version project seamlessly, usually some modifications are needed

mrproliu: I have checked the GO 1.19.2, It works well.
image

wu-sheng: Let's add 1.19 in the build matrix, and only declare we support these two major versions.

wu-sheng (dismissed):

wu-sheng (approved):

PR #121, opened 2022-11-15
Support transmit Span Attached Event protocol data.

mrproliu:


wu-sheng:
Please update the docs as well.

wu-sheng (approved):

PR #123, opened 2022-11-28
Adding kafka forwarder plugins

Arunkumarlnr:
We have added kafka forwarder plugins for event, meters, metrics, profiling, management and jvm.


wu-sheng:
Could you provide your use case?

plugins/forwarder/forwarder_repository.go
```diff
@@ -56,14 +58,16 @@ func RegisterForwarderPlugins() {
 		new(grpc_nativecds.Forwarder),
 		new(grpc_nativeevent.Forwarder),
 		new(grpc_nativejvm.Forwarder),
-		new(grpc_nativeclr.Forwarder),
-		new(grpc_nativeprocess.Forwarder),
-		new(grpc_nativeebpfprofiling.Forwarder),
```
mrproliu, line 61: Why did you need to delete the existing forwarders?

plugins/forwarder/kafka/nativemanagement/forwarder.go
```diff
@@ -0,0 +1,110 @@
...
+
+const (
+	Name     = "nativemanagement-kafka-forwarder"
+	ShowName = "Native Management Kafka Forwarder"
+)
+
+// empty struct for set
+type void struct{}
+
+var empty void
```
mrproliu, line 42: I think these fields and struct are not used, why do you declare in there?

plugins/forwarder/kafka/nativemeter/forwarder.go
```diff
@@ -0,0 +1,115 @@
...
+		rawdata, ok := proto.Marshal(data.Meter)
+		if ok != nil {
+			return ok
+		}
+		message = append(message, &sarama.ProducerMessage{
+			Topic: f.Topic,
+			Value: sarama.ByteEncoder(rawdata),
+		})
+	}
+	return f.producer.SendMessages(message)
```
mrproliu, line 102: The meter data needs to be sent to the consistent backend address. Please follow the native gRPC Meter forwarder to get more information.

test/e2e/base/consumer/target/classes/application.yml
```diff
@@ -0,0 +1,24 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
```
mrproliu, line 5: Please don't add the files under the `target` directory.

Arunkumarlnr:
@wu-sheng 
We wanted to send all of our data through Kafka from satellite component. Previously we had support for sending only logs now we have added for all others also.

mrproliu (commented):
Thanks for your contribution, please complete the following steps:
1. Fix the comments.
2. Using `make gen-docs` to generate the documentation.
3. You need to create an E2E to verify the Kafka forwards are working well.

plugins/forwarder/kafka/nativemanagement/forwarder.go
Arunkumarlnr: > Thanks for your contribution, please complete the following steps:
> 
> 1. Fix the comments.
> 2. Using `make gen-docs` to generate the documentation.
> 3. You need to create an E2E to verify the Kafka forwards are working well.

Sure will check and fix all the comments added..

wu-sheng:
Another thing matters, please provide the sizes of binary before and after these new forwarders. 
Also, we need a license update.

wu-sheng:
> @wu-sheng We wanted to send all of our data through Kafka from satellite component. Previously we had support for sending only logs now we have added for all others also.

Log's Kafka forward doesn't mean anything actually. It is a legacy feature. In the very beginning of Satellite, the agent doesn't provide log reporter, and Satellite did instead.

plugins/forwarder/kafka/nativemeter/forwarder.go
mayurvaid-redvest: @mrproliu  , Can you please elaborate a bit more on this as here we are sending to kafka and and then oap reads from the same topic

wu-sheng: Kafka has the partitioning mechanism. Some metrics like meter should always forward to the consistent partition

mayurvaid-redvest: Thanks @wu-sheng  , We will check this

PR #128, opened 2023-03-14
Kafka forwards

zbw911:
This branch provides several forwarders such as native-tracing-kafka-forwarder, native-management-kafka-forwarder, native-jvm-kafka-forwarder, and native-meter-kafka-forwarder. However, native-profile-kafka-forwarder has not been implemented yet.

1,Use case: Since the .NET Core Agent has not implemented a Kafka reporter, you can use Skywalking Satellite as a proxy to write to Kafka. However, since the OAP Kafka-fetcher-plugin has not implemented a CLR handler, the native-clr-kafka-forwarder has not been implemented in this release.

2,For most scenarios, enabling only native-tracing-kafka-forwarder can meet the needs of most business scenarios, and other requests with relatively small volume can be forwarded using a gRPC client.

3,I have already used it in production environment, it can support applications written in Python, .NET Core, and Java.

### DIFF
--- a/plugins/forwarder/kafka/nativetracing/forwarder.go
+++ b/plugins/forwarder/kafka/nativetracing/forwarder.go
@@ -25,7 +25,6 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	"github.com/apache/skywalking-satellite/internal/pkg/config"
-	"github.com/apache/skywalking-satellite/internal/pkg/log"
 	"github.com/apache/skywalking-satellite/internal/satellite/event"
 
 	v3 "skywalking.apache.org/repo/goapi/collect/language/agent/v3"
@@ -93,14 +92,16 @@ func (f *Forwarder) Forward(batch event.BatchEvents) error {
 				Value: sarama.ByteEncoder(data.Segment),
 			})
 		case *v1.SniffData_SpanAttachedEvent:
-			// SniffData_SpanAttachedEvent is from ebpf agent, skywalking-rover project.
-			// You could find it here,
-			// https://github.com/apache/skywalking-data-collect-protocol/blob/0da9c8b3e111fb51c9f8854cae16d4519462ecfe
-			// /language-agent/Tracing.proto#L244
-			// ref: https://github.com/apache/skywalking-satellite/pull/128#discussion_r1136909393
-			log.Logger.WithField("pipe", f.PipeName).Warnf("native-tracing-kafka-forwarder " +
-				"does not support messages of type SpanAttachedEvent and has discarded them." +
-				" Please choose native-tracing-grpc-forwarder as a replacement.")
+			spanAttachedEvent := &v3.SpanAttachedEvent{}
+			err := proto.Unmarshal(data.SpanAttachedEvent, spanAttachedEvent)
+			if err != nil {
+				return err
+			}
+			message = append(message, &sarama.ProducerMessage{
+				Topic: f.Topic,
+				Key:   sarama.StringEncoder(spanAttachedEvent.TraceContext.GetTraceSegmentId()),
+				Value: sarama.ByteEncoder(data.SpanAttachedEvent),
+			})
 		default:
 			continue
 		}


wu-sheng:
I want to argue the necessarily of this. If you need a Kafka as buffer and transparent layer between agents and OAP, why do you cost resources on satellite?

wu-sheng:
Directly setting up a Kafka sender at the agent side would be much better to maintain this kind of forwarder by satellite.

wu-sheng:
About the existing Kafka forward of logs, we explained why it is there at https://github.com/apache/skywalking-satellite/pull/123#issuecomment-1328990817

Anyway, @mrproliu, do you want to accept this? Or it is time we remove the Kafka forward(log) permanently in case of further confusion?

zbw911:
> 
Currently, I have performed gray testing on .NET, Java, and Python in the production environment. I am using Kafka directly for Java and Python, but the .NET agent does not support the use of Kafka. Therefore, I am using Satellite as a proxy to forward requests to Kafka for unified processing. Additionally, Kafka is necessary as a buffer, as I recently encountered a delay of approximately 1.8 billion in data due to an issue with ES, but the data was not lost. When using GRPC to receive OAP alarms, data loss also occurred. Kafka itself has value as a buffer, and there is a need for a means of converting GRPC to Kafka. When I discovered the Satellite project, I realized it was perfect for this type of scenario!

mrproliu:
At present, I think we can accept him. I checked the current official agents, they are basically only supported by the gRPC protocol. The satellite itself can be used as a protocol conversion, which may be an option for agents that need to be modified on a large scale. 
If all Agents provide Kafka's agreement later, I think it can be deleted in satellite.

wu-sheng:
Notice, I am not arguing for Kafka VS gRPC. The key point is why don't implement Kafka sender on .net agent side like Java and Python, it has better performance than using Satellite as an extra node.

wu-sheng:
> At present, I think we can accept him. I checked the current official agents, they are basically only supported by the gRPC protocol. The satellite itself can be used as a protocol conversion, which may be an option for agents that need to be modified on a large scale. If all Agents provide Kafka's agreement later, I think it can be deleted in satellite.

OK, if that is the bar, I think we will never reach :)

wu-sheng:
Please fix the CI and more important, we require a Kafka involved e2e to verify this feature. @zbw911

test/e2e/base/base-compose-kafka.yml
```diff
@@ -0,0 +1,115 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+version: '2.1'
+
+services:
```
mrproliu, line 18: The `base-compose.yml` is enough, please add the service there.

test/e2e/case/native-protocols-kafka/e2e.yaml
```diff
@@ -0,0 +1,123 @@
...
+    # native event: event list
+    - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql event list --service-name e2e-service-provider --instance-name provider1
+      expected: expected/event-list.yml
+
+    # native log: logs list
+    - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql logs list --service-name e2e-service-provider --trace-id=$(swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql trace ls|grep -A 5 '/info'|tail -n1|awk -F ' ' '{print $2}')
+      expected: expected/logs-list.yml
+
+    # native profile: create task
+    - query: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql profiling trace create --service-name e2e-service-provider --endpoint-name POST:/info --start-time=$((($(date +%s)+5)*1000)) --duration=1 --min-duration-threshold=0 --dump-period=10 --max-sampling-count=9
```
mrproliu, line 87: The profile and CDS protocol is not implemented, so why do we need to check this? Please check it locally and remove the unnecessary expected files.

mrproliu (changes requested):
Please fix the reviews and CI.

mrproliu:
There are still a lot of unprocessed previews here. Please handle them all.

plugins/forwarder/kafka/nativetracing/forwarder.go
```diff
@@ -0,0 +1,121 @@
...
+				return err
+			}
+
+			message = append(message, &sarama.ProducerMessage{
+				Topic: f.Topic,
+				Key:   sarama.StringEncoder(segmentObject.GetTraceSegmentId()),
+				Value: sarama.ByteEncoder(data.Segment),
+			})
+		case *v1.SniffData_SpanAttachedEvent:
+			message = append(message, &sarama.ProducerMessage{
```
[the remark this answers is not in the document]
zbw911: Today I was looking for the entity of SniffData_SpanAttachedEvent corresponding to the code of a Java agent, but I couldn't find it. I also tried to debug the project with breakpoints, but it didn't go into this part of the code. Can we ignore this part of the code for now, or could you give me some hints on how to locate it? Thank you.   @mrproliu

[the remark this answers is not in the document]
wu-sheng: `SniffData_SpanAttachedEvent` is from ebpf agent, skywalking-rover project.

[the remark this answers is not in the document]
wu-sheng: You could find it here, https://github.com/apache/skywalking-data-collect-protocol/blob/0da9c8b3e111fb51c9f8854cae16d4519462ecfe/language-agent/Tracing.proto#L244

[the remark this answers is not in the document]
zbw911: I carefully looked into it and we might not be able to add the SpanAttachedEvent feature to OAP anytime soon. So, for now, we can simply warn users that the current version of Kafka  forwarder  does not support this feature. It's not the best solution, but it's the only option we have for now.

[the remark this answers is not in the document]
wu-sheng: What blocks you? I think this is very easy to add. It is just a part of the trace. Routing like a segment should be enough.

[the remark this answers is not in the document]
zbw911: I am not familiar with Rover and  cannot conduct testing. If I were to proceed without proper knowledge, it could negatively impact the quality.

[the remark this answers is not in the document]
wu-sheng: You don't need to understand all. You just need to copy and change a case of it.
Take this, https://github.com/apache/skywalking/blob/2e04b069bfba24b1233dd66588217d08bdde7710/.github/workflows/skywalking.yaml#L523-L524.

Verify relative traces, but may be not real attached events.

https://github.com/apache/skywalking/blob/master/test/e2e-v2/cases/profiling/ebpf/network/network-cases.yaml#L143-L156

@mrproliu What do you want to add?

configs/satellite_config_kafka.yaml
```diff
@@ -0,0 +1,605 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# The logger configuration.
+logger:
```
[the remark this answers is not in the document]
mrproliu: Why close this? please delete this file, if you want to E2E test with this file, please move it to the test directory.

.github/workflows/build-and-test.yaml
```diff
@@ -21,6 +21,7 @@ on:
   push:
     branches:
       - main
+      - kafka-forwards
```
mrproliu, line 24: This branch does not exist. please delete it.

.github/workflows/e2e-istio.yaml
```diff
@@ -21,6 +21,7 @@ on:
   push:
     branches:
       - main
+      - kafka-forwards
```
mrproliu, line 24: Should be deleted too.

.github/workflows/e2e-native.yaml
```diff
@@ -20,6 +20,7 @@ on:
   push:
     branches:
       - main
+      - kafka-forwards
```
mrproliu, line 23: Should be deleted too.

.github/workflows/publish-docker.yaml
```diff
@@ -20,6 +20,7 @@ on:
   push:
     branches:
       - main
+      - kafka-forwards
```
mrproliu, line 23: Should be deleted too.

plugins/forwarder/kafka/nativetracing/forwarder.go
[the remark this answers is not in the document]
wu-sheng: This is not resolved.

docs/en/setup/plugins/forwarder_native-meter-kafka-forwarder.md
```diff
@@ -0,0 +1,15 @@
...
+This is a synchronization Kafka forwarder with the SkyWalking native meter protocol.
+## DefaultConfig
+```yaml
+# The remote topic. 
+topic: "skywalking-meters"
+```
+## Configuration
+|Name|Type|Description|
+|----|----|-----------|
+| routing_rule_lru_cache_size | int |  |
```
mrproliu, line 12: Seem like you didn't run the `make gen-docs` again.

plugins/forwarder/kafka/nativelog/forwarder.go
```diff
@@ -80,10 +82,16 @@ func (f *Forwarder) Forward(batch event.BatchEvents) error {
 		if !ok {
 			continue
 		}
-		for _, l := range data.LogList.Logs {
+		for _, logData := range data.LogList.Logs {
+			logdata := &v3.LogData{}
+			err := proto.Unmarshal(logData, logdata)
```
mrproliu, line 87: BTW: Have you ever tested the CPU consumption of `Unmarshal` every time you transfer data? We found that the need for meaningless encoding and decoding in each transmission will consume a lot of CPU resources, so the byte array will be transmitted directly.

plugins/forwarder/kafka/nativemeter/forwarder.go
```diff
@@ -0,0 +1,125 @@
...
+const (
+	Name     = "native-meter-kafka-forwarder"
+	ShowName = "Native Meter Kafka Forwarder"
+)
+
+type Forwarder struct {
+	config.CommonFields
+	RoutingRuleLRUCacheSize int `mapstructure:"routing_rule_lru_cache_size"`
+	// The TTL of the LRU cache size for hosting routine rules of service instance.
+	RoutingRuleLRUCacheTTL int    `mapstructure:"routing_rule_lru_cache_ttl"`
```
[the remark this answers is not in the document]
mrproliu: Please fix it.

plugins/forwarder/kafka/nativelog/forwarder.go
zbw911: Yes, there is indeed some impact on performance, although I haven't delved deeply into how much. In a previous version, I generated entities with only a few necessary fields for deserialization through a proto file. However, I found that I needed to maintain another entity. As a result, I deleted the generated files. Another ultimate solution, as you suggested, is to unpack the binary stream and obtain the corresponding fields based on the delimiter. We'll leave this as an optimization point for now, as it is sufficient for our current needs. However, as more applications are integrated later, we may optimize according to these solutions.

wu-sheng: I don't think we should loss performance due to add Kafka forward.
That isn't how things work. 
Please keep what we do for now, until open process mechanism discussion, and have new resolution.

plugins/forwarder/kafka/nativetracing/forwarder.go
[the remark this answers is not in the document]
zbw911: I'm not quite sure if it's like this. If it is, I'll continue tomorrow when I have time. Thank you. 
https://github.com/zbw911/skywalking-satellite/blob/e86c410f1d06961f9b4be0dd26131772245f1ed2/plugins/forwarder/kafka/nativetracing/forwarder.go#L94

plugins/forwarder/kafka/nativelog/forwarder.go
zbw911: I agree with your point of view. Let's make it run first and optimize it later. Then, let's mark it as resolved.

wu-sheng: @mrproliu As a log forwarder, what it is the current way(gRPC) to process this? Is this consistent with that?

mrproliu: When the current log uses gRPC to receive messages, it does not `Unmarshal`, it directly reads binary data for processing. When using gRPC to send data, directly send the current stream for ten days, because it does not need to do anything based on metadata, so he does not need to use `Umarshal` to do anything.
So, gRPC is fine.

wu-sheng: Then, we need to make Kafka follow the same way. If the gRPC forwards to the same channel, the Kafka forward should to the same partition.
@mrproliu Let's make sure this Kafka forward is always consistent with gRPC routing.

mrproliu: For now, the gRPC didn't forward the same instance to the upstream, just make sure one stream send to the one upstream. 
If Kafka wants to forward to the same partition, then, just following what I say, decode the first log in the list, which is an element of the channel, don't decode them all.

plugins/forwarder/kafka/nativetracing/forwarder.go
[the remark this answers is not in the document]
zbw911: > Why does the attached event hasn't a key of the message? Also, will you create a PR for handling the attached event in the OAP?

A TracingType currently covers two types of business requests, which has led to the current if-else problem. The setting of SniffType needs to be further detailed and each type should be a type of business.

Types should be defined as more specific types rather than classifications. SniffType should be mapped to the service layer in proto.

Finally, my solution is to add SpanAttachedEventType = 17 in https://github.com/apache/skywalking-goapi/blob/main/satellite/data/v1/SniffData.proto#L47.

If possible, I will branch the goapi project, and then modify the skywalking-satellite project.

@mrproliu  

-------
v1.SniffData_SpanAttachedEvent->SpanAttachedEventReportService：  https://github.com/apache/skywalking-data-collect-protocol/blob/0da9c8b3e111fb51c9f8854cae16d4519462ecfe/language-agent/Tracing.proto#L233-L237

v1.SniffData_Segment->TraceSegmentReportService：  https://github.com/apache/skywalking-data-collect-protocol/blob/0da9c8b3e111fb51c9f8854cae16d4519462ecfe/language-agent/Tracing.proto#L33-L46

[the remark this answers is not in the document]
mrproliu: Why do you think the `if-else` is a problem? They all belong to the tracing data. For this review, the question is the Attached span hasn't a message key, not another question.

plugins/forwarder/kafka/nativelog/forwarder.go
mrproliu: Why close it? this issue seems not been updated. please follow my comment to fix it.

plugins/forwarder/kafka/nativetracing/forwarder.go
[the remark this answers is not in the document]
zbw911: The existing OAP server cannot parse SpanAttachedEvent (https://github.com/apache/skywalking/blob/2e04b069bfba24b1233dd66588217d08bdde7710/oap-server/server-fetcher-plugin/kafka-fetcher-plugin/src/main/java/org/apache/skywalking/oap/server/analyzer/agent/kafka/provider/handler/TraceSegmentHandler.java#L71-L86).

Parsing in the current mode may look like the following:
``` 
try{
        SegmentObject segment = SegmentObject.parseFrom(record.value().get());
        segmentParserService.send(segment);
}
catch(InvalidProtocolBufferException e)
{
      try{
          SpanAttachedEvent spanAttachedEvent = SpanAttachedEvent.parseFrom(record.value().get());
          //.................
      }
      catch(InvalidProtocolBufferException e )
      {
      // ...........
      }
}
```

Making changes to the above code is somewhat difficult.

Therefore, a warning message should be added to the tracing forwarder, explicitly indicating that the current Kafka mode does not support SpanAttachedEvent, so that users can be aware of this limitation and take appropriate action.

Thus, instead of adding a key to SpanAttachedEvent, we should explicitly inform the user that the current Kafka mode does not support this event to prevent misuse.

If necessary, we can increase the log level to error.

Additionally, I suggest refining the type in goapi as a possible future option.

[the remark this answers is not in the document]
zbw911: I think the best approach for now is to temporarily abandon support for *v1.SniffData_SpanAttachedEvent in Kafka mode.

plugins/forwarder/kafka/nativelog/forwarder.go
zbw911: Ok, I known.

plugins/forwarder/kafka/nativetracing/forwarder.go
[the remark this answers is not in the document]
wu-sheng: Are you really reading the protocol? Read more carefully, rather than guessing.

https://github.com/apache/skywalking-data-collect-protocol/blob/0da9c8b3e111fb51c9f8854cae16d4519462ecfe/language-agent/Tracing.proto#L233-L237

What you thought is actually impossible from gRPC stand point.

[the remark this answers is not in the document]
wu-sheng: All protocols are only getting accepted if the backend supports it. 
Be clear about the status, and be patient in reading projects and docs. 

You are proposing a big change. From a maintainer perspective, it is already not easy to follow your first contribution as huge, you should get familiar with the project more rather than keeping saying `best approach` or `should`.

Again and again, we have given you the e2e to simulate how to use ebpf agent and all these protocols. You at least should follow them and try them.

What you replied makes me concerned you are only pushing this in hurry, rather than really want to understand what is the case.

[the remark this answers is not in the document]
wu-sheng: @mrproliu Please be very careful about all the codes this PR is supposed to change. I am super concerned about it lacking fully testing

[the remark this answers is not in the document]
mrproliu: @zbw911 please finish this issue first, then click the `resolve conversation` button. 

> The existing OAP server cannot parse SpanAttachedEvent

Back to my first comment on this conversation, you should also implement the relevant functions on the OAP side.
You have already implemented the management Kafka forwarder, so you should know there are also multiple data structures in that service. In that implementation, you already use the `different message key prefixes` to identify them, you can do this in the same way. 
You should read the OAP code related to the management Kafka receiver: https://github.com/apache/skywalking/blob/master/oap-server/server-fetcher-plugin/kafka-fetcher-plugin/src/main/java/org/apache/skywalking/oap/server/analyzer/agent/kafka/provider/handler/ServiceManagementHandler.java#L61-L71

plugins/forwarder/kafka/nativetracing/forwarder.go
```diff
@@ -0,0 +1,121 @@
...
+				Key:   sarama.StringEncoder(segmentObject.GetTraceSegmentId()),
+				Value: sarama.ByteEncoder(data.Segment),
+			})
+		case *v1.SniffData_SpanAttachedEvent:
+			// SniffData_SpanAttachedEvent is from ebpf agent, skywalking-rover project.
+			// You could find it here,
+			// https://github.com/apache/skywalking-data-collect-protocol/blob/0da9c8b3e111fb51c9f8854cae16d4519462ecfe
+			// /language-agent/Tracing.proto#L244
+			// ref: https://github.com/apache/skywalking-satellite/pull/128#discussion_r1136909393
+			log.Logger.WithField("pipe", f.PipeName).Warnf("native-tracing-kafka-forwarder " +
```
mrproliu, line 101: Why there still not implemented?

test/e2e/case/native-protocols-kafka/satellite_config_kafka.yaml
```diff
@@ -0,0 +1,597 @@
...
+    client_name: ${SATELLITE_TELEMETRY_METRICS_SERVICE_CLIENT_NAME:grpc-client}
+    # The interval second for sending metrics
+    interval: ${SATELLITE_TELEMETRY_METRICS_SERVICE_INTERVAL:10}
+    # The prefix of telemetry metric name
+    metric_prefix: ${SATELLITE_TELEMETRY_METRICS_SERVICE_METRIC_PREFIX:sw_stl_}
+
+# The sharing plugins referenced by the specific plugins in the different pipes.
+sharing:
+  clients:
+    - plugin_name: "grpc-client"
```
mrproliu, line 55: This client is unnecessary, please remove.

test/e2e/case/native-protocols-kafka/satellite_config_kafka.yaml
```diff
@@ -0,0 +1,597 @@
...
+      flush_time: ${SATELLITE_TRACINGPIPE_SENDER_FLUSH_TIME:1000}
+      # The maximum buffer elements.
+      max_buffer_size: ${SATELLITE_TRACINGPIPE_SENDER_MAX_BUFFER_SIZE:200}
+      # The minimum flush elements.
+      min_flush_events: ${SATELLITE_TRACINGPIPE_SENDER_MIN_FLUSH_EVENTS:1}
+      client_name: kafka-client
+      forwarders:
+        - plugin_name: native-tracing-kafka-forwarder
+          topic: ${SATELLITE_FORWARDER_KAFKA_TOPIC_SEGMENT:skywalking-segments}
+  - common_config:
```
mrproliu, line 236: There still have many not using pipelines, please remove them, to prevent abuse due to unknown circumstances. Such as profile/CDS/event/CLR/ALS pipe, etc.

plugins/forwarder/kafka/nativetracing/forwarder.go
wu-sheng: > Can you confirm whether e2e can be run on WSL?

Using Linux and macOS would be much much better than WSL on Windows. That is pain.

zbw911: Do you really not support WSL? If that's the case, then I'll have to switch to that Mac. PS: I'm not very used to using Mac! :(

mrproliu: It should be works well with windows. Could you provide more detail about the error?

wu-sheng: We don't test on WSL. No one in the maintainer team is using Windows. That is the point.  Linux is very easy to set up. You could run from there on a Windows.

zbw911: > It should be works well with windows. Could you provide more detail about the error?

the example of skywalking-infra-e2e:

1，windows CMD:
```
F:\github\apache\skywalking-infra-e2e\bin\windows>.\e2e.exe  run -c ../../examples/compose/e2e.yaml
INFO load the e2e config successfully             
: invalid reference format: no such image: apache/skywalking-oap-server:8.5.0-es6
INFO deleting docker compose cluster...           
INFO cleanup part finished successfully           
ERROR Local Docker compose exited abnormally whilst running docker-compose.exe: [--env-file F:\github\apache\skywalking-infra-e2e\examples\compose\env up -d]. exit status 1 
```

2，windows PowerShell:
```
PS F:\github\apache\skywalking-infra-e2e\bin\windows> .\e2e.exe  run -c ../../examples/compose/e2e.yaml
INFO load the e2e config successfully
: invalid reference format: no such image: apache/skywalking-ui:8.5.0
INFO deleting docker compose cluster...
INFO cleanup part finished successfully
ERROR Local Docker compose exited abnormally whilst running docker-compose.exe: [--env-file F:\github\apache\skywalking-infra-e2e\examples\compose\env up -d]. exit status 1
```

3，WSL 2:
```
root@DESKTOP-P1D7M3U:/mnt/f/github/apache/skywalking-infra-e2e/bin/linux# e2e run -c ../../examples/compose/e2e.yaml
INFO load the e2e config successfully
: invalid reference format: no such image: apache/skywalking-ui:8.5.0
INFO deleting docker compose cluster...
INFO cleanup part finished successfully
ERROR Local Docker compose exited abnormally whilst running docker-compose: [--env-file /mnt/f/github/apache/skywalking-infra-e2e/examples/compose/env up -d]. exit status 1
```



docker-compose :
```
root@DESKTOP-P1D7M3U:/mnt/f/github/apache/skywalking-infra-e2e/examples/compose# docker-compose --env-file=./env  up -d
[+] Running 2/2
 ⠿ Container compose-oap-1  Healthy                                                                                                                                                                                                                                  11.6s 
 ⠿ Container compose-ui-1   Started  
```

I have been able to successfully start using the docker-compose command on both Windows and WSL. However, I have never been successful in using the e2e command directly.

zbw911: > We don't test on WSL. No one in the maintainer team is using Windows. That is the point. Linux is very easy to set up. You could run from there on a Windows.

I'm going to try installing a Linux desktop on Hyper-V.

zbw911: > It should be works well with windows.

The issue was caused by LF and CRLF. To fix this,   add .sh, env, and mvnw to the .gitattributes file in Git. This will prevent them from being converted to CRLF (\r\n) when cloned.   then use dos2unix or Goland (file->file properties->Line separators->LF unix and macOS(\n)) to resolve the issue. Now it works well on WSL. This is a common problem that Windows users may encounter.

wu-sheng: I think ebpf agent(AttachedEvent) is highly relying on Linux. I even doubt MacOS would work.
Don't try to depend on one Platforma, sometimes, it just doesn't work from one case to another.

zbw911: @mrproliu  Help!!!
I've been trying to do e2e testing for several days now, trying various methods such as using swctl to query with a single sentence, and also modifying service-instance.yml to adapt to the return value, but when it comes to the process list step, it always returns [].

I have not been able to successfully execute the test, and the whole process has been stuck in this part, making progress extremely difficult.

I hope you can help me figure out how to make this part run smoothly, and what kind of environment or tools I need to use. I have tried almost all the methods that I can try now.

My usage is Docker Desktop, and the instance is running normally from Containers.

Below is the original error message from the main branch:
----------
```
  ERROR failed to verify the output: swctl --display yaml --base-url=http://${oap_host}:${oap_12800}/graphql instance list --service-name=service, error:
mismatch (-want +got):
  []interface{}{
-       map[interface{}]interface{}{
-               string("attributes"):   []interface{}{},
-               string("id"):           string("c2VydmljZQ==.1_dGVzdA=="),
-               string("instanceuuid"): string("c2VydmljZQ==.1_NTFjOGZmOTBjYmE2MTFlZGEwYjIwMjQyYWMxOTAwMDRAMTcyLjI1LjAuNA=="),
-               string("language"):     string("UNKNOWN"),
-               string("name"):         string("test"),
-       },
+       map[interface{}]interface{}{
+               string("attributes"): []interface{}{
+                       map[interface{}]interface{}{string("name"): string("Process No."), string("value"): string("1")},
+                       map[interface{}]interface{}{string("name"): string("hostname"), string("value"): string("e7786ddc74c7")},
+                       map[interface{}]interface{}{string("name"): string("OS Name"), string("value"): string("linux")},
+                       map[interface{}]interface{}{string("name"): string("ipv4s"), string("value"): string("172.25.0.4")},
+               },
+               string("id"):           string("c2VydmljZQ==.1_NTFjOGZmOTBjYmE2MTFlZGEwYjIwMjQyYWMxOTAwMDRAMTcyLjI1LjAuNA=="),
+               string("instanceuuid"): string("c2VydmljZQ==.1_NTFjOGZmOTBjYmE2MTFlZGEwYjIwMjQyYWMxOTAwMDRAMTcyLjI1LjAuNA=="),
+               string("language"):     string("GO"),
+               string("name"):         string("51c8ff90cba611eda0b20242ac190004@172.25.0.4"),
+       },
  }
```

wu-sheng:
No update for months. Closing for now.

PR #131, opened 2023-04-02
Enhance debuggability to help identify issues in the environment

mrproliu:
At present, Satellite does not have much ability to check for problems. When problems occur after deployment, it is difficult to identify the cause of the problem. Therefore, monitoring capabilities and some logs have been strengthened here.

### DIFF
--- a/docs/en/setup/examples/feature/telemetry-exporter/README.md
+++ b/docs/en/setup/examples/feature/telemetry-exporter/README.md
@@ -59,7 +59,7 @@ telemetry:
     metric_prefix: ${SATELLITE_TELEMETRY_METRICS_SERVICE_METRIC_PREFIX:sw_stl_}
 ```
 
-## Pprof
+## pprof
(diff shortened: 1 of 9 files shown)


docs/en/setup/examples/feature/telemetry-exporter/README.md
```diff
@@ -55,4 +57,20 @@ telemetry:
     interval: ${SATELLITE_TELEMETRY_METRICS_SERVICE_INTERVAL:10}
     # The prefix of telemetry metric name
     metric_prefix: ${SATELLITE_TELEMETRY_METRICS_SERVICE_METRIC_PREFIX:sw_stl_}
+```
+
+## Pprof
```
wu-sheng, line 62:
```suggestion
## pprof
```

I think we usually don't put upper case for this project?

wu-sheng (approved):

PR #132, opened 2023-05-11
Add timeout config for the gRPC client

mrproliu:


wu-sheng (approved):

wu-sheng:
Let's update showcase and demo to see whether the bug is fixed properly.

PR #134, opened 2023-06-01
Bump up libssl to fix CVE

kezhenxu94:
@mrproliu I think we don't need to use the base image `debian` , can we use something like `alpine`?


mrproliu:
> can we use something like alpine?

Yes, I could be updated to `alpine`

kezhenxu94:
> > can we use something like alpine?
> 
> Yes, I could be updated to `alpine`

If you can do that, let's close this PR

PR #155, opened 2024-09-06
Allow manually rebuild docker images

kezhenxu94:


.github/workflows/publish-docker.yaml
```diff
@@ -20,6 +20,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
```
wu-sheng, line 23: Don't we need a version parameter to manually build?

kezhenxu94: > Don't we need a version parameter to manually build?

No. We would either rebuild the latest master branch or need code changes in latest master. Rebuilding an old commit doesn't make sense.

wu-sheng: I see. You want to rebuild the latest for local branch? Is that the purpose?

kezhenxu94: yes

wu-sheng: So, this would not work for fork repo, mostly it is for committers' branch.

wu-sheng (approved):

kezhenxu94:
We should consider upgrade the Istio version in the tests, 1.7 ~ 1.9 looks really outdated

PR #156, opened 2024-11-08
Support async-profiler grpc interface

zhengziyi0117:


wu-sheng:
You need to add the new case into gha control file.

plugins/client/grpc/client_config.go
```diff
@@ -68,14 +68,18 @@ func (c *Client) loadConfig() (*[]grpc.DialOption, error) {
 		if authHeader != nil {
 			ctx = metadata.NewOutgoingContext(ctx, authHeader)
 		}
+		supportBidirectionalStream := false
+		if b := ctx.Value("BidirectionalStream"); b != nil {
```
mrproliu, line 72: Please make the key as a const.

plugins/forwarder/api/forwarder.go
```diff
@@ -34,7 +35,7 @@ type Forwarder interface {
 	// Forward the batch events to the external services, such as Kafka MQ and SkyWalking OAP cluster.
 	Forward(batch event.BatchEvents) error
 	// SyncForward the single event to the external service with sync forward
-	SyncForward(event *v1.SniffData) (*v1.SniffData, error)
+	SyncForward(event *v1.SniffData) (*v1.SniffData, grpc.ClientStream, error)
```
mrproliu, line 38: Please add a comment about when the user need to use the client stream.

plugins/forwarder/grpc/nativeasyncprofiler/forwarder.go
```diff
@@ -0,0 +1,108 @@
...
+		query := requestData.AsyncProfilerTaskCommandQuery
+		commands, err := f.profilingClient.GetAsyncProfilerTaskCommands(context.Background(), query)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		return &v1.SniffData{Data: &v1.SniffData_Commands{Commands: commands}}, nil, nil
+	case *v1.SniffData_AsyncProfilerData:
+		// metadata
+		ctx := context.WithValue(context.Background(), "BidirectionalStream", true)
```
mrproliu, line 70: This key should reference the const which defined in the grpc client  package.

mrproliu (commented):
Thanks, please generate the documentation by "make gen-docs" command.

mrproliu (approved):

PR #174, opened 2025-02-05
chore: move license generation to release phase

kezhenxu94:
Instead of generating LICENSE in each PR, moving this to release phase can help dependencies upgrade to be more smoothly, this PR also add a GitHub Actions step to check incompatible license as early as possible, to avoid block when doing a release.


wu-sheng:
So, from this, PR CI only checks the dependencies compatibility, but don't force the license comparison, am I understanding this correctly?

kezhenxu94:
> So, from this, PR CI only checks the dependencies compatibility, but don't force the license comparison, am I understanding this correctly?

Yes, with this PR, LICENSE file and licenses/ folder will not be compared or checked, we just make sure the licenses are compatible so we can easily generate LICENSE and licenses/ without having to revert incompatible dependencies till releasing.

docs/en/guides/contribution/How-to-release.md
```diff
@@ -6,7 +6,10 @@ This documentation guides the release manager to release the SkyWalking Satellit
 
 1. Close(if finished, or move to next milestone otherwise) all issues in the current milestone from [skywalking-satellite](https://github.com/apache/skywalking-satellite/milestones) and [skywalking](https://github.com/apache/skywalking/milestones), create a new milestone if needed.
 2. Update [CHANGES.md](../../../../CHANGES.md).
-
+3. Generate LICENSE file in [`dist/LICENSE`](../../../../dist/LICENSE) and copy dependencies' licneses to [`dist/licenses`](../../../../dist/licenses/) folder, this can be done by running the following command in the project root:
+```shell
+license-eye dep resolve --summary dist/LICENSE.tpl -o dist/licenses
```
wu-sheng, line 11: Should we consider to locking eyes version?
I am asking this, because I want this step in oap and Java agent too.

kezhenxu94: Added instruction to install specific version of license-eye

wu-sheng (approved):

wu-sheng:
Could you apply this to oap and java agent as well? It should lower the 1st contributor bar.

